PR #5, opened 2026-06-15
Fix install.sh self-recursion hang and false-success on flat release tarball

### LINKED ISSUE #4: install-graff.sh hangs forever (infinite self-recursion) + installs nothing (wrong tarball layout)
## Summary

`curl -fsSL https://codegraff.com/install-graff.sh | sh` hangs forever for a fresh install. After working around the hang, the script reports success but **no binary is actually installed**. Two separate bugs.

Environment: macOS (darwin-arm64), zsh.

## Bug 1 — infinite self-recursion (the hang)

`https://codegraff.com/install-graff.sh` serves the installer itself. In the "graff companion suite" step the script runs:

```sh
# line ~175
if curl -fsSL https://codegraff.com/install-graff.sh | sh >/dev/null 2>&1; then
```

That URL **is this same script**. The guard above it (`command -v muonry && command -v zigpatch`) fails on a fresh machine because those tools aren't installed yet, so every invocation re-enters the same branch and re-runs itself — forever, silently (`>/dev/null 2>&1`). Result: the installer never returns.

Workaround: `HARNESS_NO_GRAFF=1 sh install-graff.sh` skips the block and the install completes.

Fix: the companion-suite step should fetch the *companion* installer (muonry/zigrep/zigpatch suite), not re-curl `install-graff.sh`. As written it can never make `muonry`/`zigpatch` present, so it can only recurse.

## Bug 2 — wrong tarball layout, nothing gets installed

Even with the hang bypassed, the prebuilt-release path fails:

```
install: /var/folders/.../tmp.XXXX/graff-aarch64-macos/graff: No such file or directory
  │ download   ✓ (prebuilt release)
  │ install    ✓
  installed → /Users/<me>/bin/graff
```

`fetch_release` extracts the tarball and calls (line ~108):

```sh
place_bin "$tmpd/$stem-$target/$stem"   # expects graff-aarch64-macos/graff
```

But the actual release asset `graff-aarch64-macos.tar.gz` extracts **flat**:

```
$ tar -tzf graff-aarch64-macos.tar.gz
graff
README.md
LICENSE
```

There is no `graff-aarch64-macos/` subdirectory, so `install` gets a nonexistent path and fails. The failure is swallowed (the codesign/`|| true` flow continues) and the script still prints `installed ✓` — a false success.

Fix options: either (a) package the tarball with the expected `graff-<target>/graff` subdirectory, or (b) change `place_bin` to use the flat path (`$tmpd/$stem`), or locate the binary after extraction rather than assuming the layout. Whichever is chosen, `fetch_release` should also *verify the binary exists* before reporting `download ✓`, so a failed extract doesn't print a false success.

## Repro

```sh
curl -fsSL https://codegraff.com/install-graff.sh | sh    # hangs forever
# vs
HARNESS_NO_GRAFF=1 sh install-graff.sh                    # no hang, but no binary installed
```

## Manual install that does work

```sh
curl -fsSL https://github.com/justrach/codegraff/releases/latest/download/graff-aarch64-macos.tar.gz -o graff.tar.gz
tar -xzf graff.tar.gz
install -m 0755 graff "$HOME/bin/graff"
codesign -s - --force "$HOME/bin/graff"   # macOS
$HOME/bin/graff --version                 # -> simple-harness 0.0.1
```

pranavp311:
Fixes #4.

`curl -fsSL https://codegraff.com/install-graff.sh | sh` hung forever on a fresh machine; bypassing the hang then installed nothing while reporting success. Two independent bugs in `install.sh`, both fixed here.

## Bug 1 — infinite self-recursion (the hang)
The companion-suite step ran `curl -fsSL https://codegraff.com/install-graff.sh | sh`, but that URL serves **this same installer**. The guarding branch (`muonry`/`zigpatch` present) can never be satisfied by re-running the graff installer, so every nested invocation re-enters the step and recurses forever — the installer never returns. (A nested run can also `rm` the just-placed binary and then fail to reinstall, leaving a dangling `harness` symlink and no `graff`.)

**Fix:** replaced the recursive `curl … | sh` with a clear `skipped` message + guidance. The step never actually installed the suite from that URL, so removing the recursion loses no working behavior. Left a comment so a real suite-installer URL can be wired in later.

## Bug 2 — wrong tarball layout, false success
`fetch_release` assumed the release tarball nested the binary under `$stem-$target/` (e.g. `graff-aarch64-macos/graff`). Current release tarballs are **flat**:

```
$ tar -tzf graff-aarch64-macos.tar.gz
graff
README.md
LICENSE
```

So `install` received a nonexistent path and failed — yet the script still printed `installed`. **Fix:** locate the binary in either the flat or nested layout, and `return 1` (fall back to source build) when it isn't found, instead of reporting a false success.

## Verification (macOS, darwin-arm64)
- `bash -n install.sh` — clean.
- Fresh run with a hard 60s alarm now **terminates immediately**, downloads the prebuilt release, installs a working `graff` (+ valid `harness` symlink), and the companion step prints `skipped`:
  ```
  │ download   ✓ (prebuilt release)
  │ install    ✓
  installed → .../graff
  │ graff      skipped (install muonry + zigrep/zigpatch suite separately; ...)
  ```
- `graff --version` → `simple-harness 0.0.1`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -105,7 +105,18 @@ fetch_release() {
   done
   [ -f "$tmpd/$asset" ] || { rm -rf "$tmpd"; return 1; }
   tar -xzf "$tmpd/$asset" -C "$tmpd" 2>/dev/null || { rm -rf "$tmpd"; return 1; }
-  place_bin "$tmpd/$stem-$target/$stem"
+  # Current release tarballs are flat ($stem at the root); older ones nested
+  # the binary under a $stem-$target/ dir. Accept either, and fail (so we fall
+  # back to source) instead of handing place_bin a path that doesn't exist.
+  local binsrc
+  if [ -f "$tmpd/$stem" ]; then
+    binsrc="$tmpd/$stem"
+  elif [ -f "$tmpd/$stem-$target/$stem" ]; then
+    binsrc="$tmpd/$stem-$target/$stem"
+  else
+    rm -rf "$tmpd"; return 1
+  fi
+  place_bin "$binsrc"
   if [ "$(uname -s)" = "Darwin" ]; then
     codesign -s - --force "$INSTALL_DIR/$BIN" >/dev/null 2>&1 || true
   fi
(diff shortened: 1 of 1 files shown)


pranavp311:
Closing in favor of #6, which lands the same diagnosis with a more complete fix: it re-points the companion step at the correct `codegraff.com/install.sh` route (the tools installer) instead of disabling it, adds a `find` fallback for unknown tarball layouts, and a final `[ -x "$INSTALL_DIR/$BIN" ]` fail-closed check. Thanks!

PR #9, opened 2026-06-16
Fix: drop reasoning_effort when the model rejects it (gpt-5.5)

justrach:
Cherry-picks the fix from the GUI branch. `effortApplies()` gates by provider (codegraff/deepseek), but those gateways serve non-reasoning chat/completions models — gpt-5.5 returns `Function tools with reasoning_effort are not supported ... use /v1/responses`, breaking the turn. Mirror the `cap_new` learned-rejection pattern: on that error, drop reasoning_effort and retry transparently; relearn on model switch. Build + fmt + tests pass.


justrach:
Redundant — the reasoning_effort crash-fix landed via #11 (commit f22dd95). Closing.

PR #32, opened 2026-06-18
gui: revert chat selection when snapshot has no conversation view

pranavp311:
Reverts the chat selection when an incoming snapshot has no conversation view, preventing the UI from getting stuck on a selection that no longer resolves.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #36, opened 2026-06-18
Fix duplicate activity keys in chat thread

pranavp311:
## Summary
- make finalized chat activity keys unique after flushed activity groups are concatenated
- add a regression test for repeated operation groups within one request

## Test plan
- bun test src/components/chat/utils/chatThread.test.ts

## Stack
Base: fix-chat-selection-revert (#32)


pranavp311:
Rebased onto the rebased `fix-chat-selection-revert` (#32). Only this PR's own commit was replayed (#32's commit detected as already-applied and skipped). Base intentionally kept on #32; will auto-retarget to `release/0.0.14` when #32 merges.

PR #37, opened 2026-06-19
Add `graff update` self-update command

pranavp311:
## What

Adds a version-checked `graff update` subcommand that upgrades the installed binary from the latest GitHub release.

- `graff update` — queries the GitHub releases API, compares the latest tag to the running version, and **skips with "already up to date" when current**. Otherwise it delegates the download/codesign/atomic binary swap to the existing `install.sh` (`curl … | sh`) rather than reimplementing that platform-specific logic. Runs the installer with `HARNESS_NO_GRAFF=1` so an update touches only `graff`, not the companion suite.
- `graff update --force` — reinstall even if already current.
- `graff update --check` — report availability only; install nothing.
- Install URL overridable via `GRAFF_INSTALL_URL` / `HARNESS_INSTALL_URL`.

Follows standard CLI self-update conventions (`deno upgrade` / `bun upgrade` / `rustup update`): explicit command, version-checked, no silent startup auto-update.

## Implementation

~99 lines in `src/main.zig`: a new `updateCommand` function plus dispatch/flag/usage wiring. Reuses existing patterns — the GET-with-User-Agent from `rawFetch`, the JSON parse from `oauthTokenPost`, and the spawn+wait from the skill installer.

## Verification

- `zig build` clean.
- `graff --help` shows the new `update` line.
- `graff update --check` (dev build) → `update available: v0.0.12… → v0.0.13`, exit 0, nothing installed.
- Version-pinned build (`-Dversion=0.0.13`) → `graff update` prints "already up to date" and does **not** spawn the installer; `--check` agrees. Exit 0.

Not exercised live: the actual `--force` install (would replace the running binary) and the offline-failure path (network couldn't be cut here; code catches fetch/parse errors → `--check` exits via `fatal`, plain `update` warns and falls through to the installer).

## Note for reviewers

The version compare is exact-string equality after stripping a leading `v` — it answers "is this exactly the latest release," not semver ordering. Matches intended usage; a true "only upgrade, never downgrade" check would need semver comparison.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -3606,7 +3606,10 @@ fn updateCommand(
 
     // Delegate download/codesign/atomic swap to install.sh. Inherit our stdio
     // so its progress (and any sudo prompt) reaches the terminal directly.
-    const cmd = try std.fmt.allocPrint(arena, "curl -fsSL {s} | HARNESS_NO_GRAFF=1 sh", .{install_url});
+    // "set -o pipefail" is required: without it, a failed "curl" is masked
+    // by the right-hand "sh" exiting 0 on EOF, and the pipeline reports
+    // success while installing nothing — silently.
+    const cmd = try std.fmt.allocPrint(arena, "set -o pipefail; curl -fsSL {s} | HARNESS_NO_GRAFF=1 sh", .{install_url});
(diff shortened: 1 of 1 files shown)


pranavp311:
Rebased onto `release/0.0.14` (2 commits, no conflicts). Base retargeted from `main` → `release/0.0.14`. Includes the `set -o pipefail` fix from the second commit.

PR #38, opened 2026-06-20
Add /bash and show current folder in prompt

yxlyx:
## Summary
- replace the proposed /pwd command with /bash <command>
- run CLI /bash through the existing bash tool path so approval/codedb guard/hooks still apply
- add GUI /bash command support with raw argument preservation and formatted stdout/stderr/exit output
- include the current folder in the startup banner and interactive model prompt

## Testing
- zig build test
- zig build
- cargo check --manifest-path gui/src-tauri/Cargo.toml *(using a temporary local graff sidecar at gui/src-tauri/binaries/graff-aarch64-apple-darwin)*
- cd gui && bun run build:deps && bun test

### DIFF
--- a/gui/src-tauri/src/runtime/simple.rs
+++ b/gui/src-tauri/src/runtime/simple.rs
@@ -403,6 +403,7 @@ impl RuntimeManager {
         Ok(vec![
             command("help", "Show available commands.", false),
             command("agent", "Show active agent.", false),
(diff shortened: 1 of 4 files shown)


justrach:
Already integrated into release/0.0.16 (commit 81d13ff). Closing as redundant against the release branch.

PR #67, opened 2026-06-21
fix(gui): auto-detect CLI provider login

justrach:
## Summary\n- Poll provider status while a CLI login flow is active\n- Auto-complete provider setup once the credential lands on disk\n- Disable the manual finish button while detection is in progress\n\n## Checks\n- cd gui && bun run build:deps && bunx tsc -b\n- cd gui/src-tauri && cargo check\n\nNote: release workflow / SDK packaging changes from the stash were already present on release/0.0.161, so this PR contains only the remaining provider setup dialog fix.

### DIFF
--- a/gui/src-tauri/src/runtime/simple.rs
+++ b/gui/src-tauri/src/runtime/simple.rs
@@ -6,6 +6,7 @@ use crate::{
     terminal::TerminalManager,
 };
 use anyhow::{Context, Result};
+use base64::Engine;
 #[cfg(target_os = "macos")]
 use std::os::unix::fs::PermissionsExt;
 use std::{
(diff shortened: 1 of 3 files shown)


justrach:
Updated this PR with GUI auth-expiry handling:\n\n- Treat locally expired Codex/Kimi OAuth tokens as not configured before starting a run.\n- Detect invalid/expired provider-auth errors from graff, show a reconnect-provider message, and drop the live session so retry starts fresh.\n\nChecks rerun:\n- cd gui/src-tauri && cargo check\n- cd gui/src-tauri && cargo test auth\n- cd gui && bun run build:deps && bunx tsc -b

Issue #124, opened 2026-07-08
Long-running sessions grow RSS unboundedly — root Agent.arena is never reset (benchmarks/memory.py can't see it)


justrach:
## Problem: current memory benchmarking can't see the bug that actually matters

`benchmarks/memory.py` measures peak RSS of a single `graff --json` invocation that receives one message on stdin and exits — that's a real, honestly-reported number (~25 MB floor), but it's structurally a **one-shot, single-turn measurement**. It can't observe anything that only shows up after N turns in the same process — which is exactly the shape of two already-identified, still-live issues.

### Confirmed, not hypothetical

Both of these are cited in a test fixture at `src/repl.zig:1561` (`renderMarkdown: too-wide table falls back to record layout`) as a findings table, but were never filed:

1. **`main.zig:5719`, root `Agent.arena` — never reset.** `fn main`'s `arena` is `init.arena.allocator()` — Zig's process-init arena, conventionally freed only at process exit. The root `Agent.arena` field (`main.zig:10885`) is set to exactly this (`main.zig:6371`), and it's the allocator behind ~150+ call sites across the whole request/tool-loop path: every streamed SSE chunk parsed via `parseFromSliceLeaky` (e.g. `assembleAnthropic`, `main.zig:14048`), every tool result, every duped string, every constructed message. There is **zero `.reset(` call anywhere in `main.zig`** (checked directly). For a one-shot invocation this never matters — the process exits before it's observable. For REPL mode, `serve` mode, or the GUI's persistent `graff --json --resume <id>` child (which lives for the life of a conversation), this accumulates permanently across every turn.
2. **`repl.zig:690` (`Model.render`) — full scrollback re-styled and re-allocated every frame.** Each `render()` call opens a fresh arena and walks the *entire* `self.history.items` to restyle every past message — welcome/input/result/assistant entries all re-rendered from scratch, every tick. Not a leak (the arena is deinit'd at the end of the function), but O(conversation length) allocation churn on every redraw — cost grows with session length the same way the arena does, just via churn instead of retention.

The existing "Memory is task-dependent" caveat in `benchmarks/README.md` attributes RSS growth solely to "how much code it reads into context." That's a real, separate, legitimate cost — but it's not what's described above. Turn-count-driven growth and context-size-driven growth are two different axes, and right now only one of them is measured or even mentioned.

## What profiling would actually catch this

[...]

PR #162, opened 2026-07-11
perf(#124): per-turn scratch arena + session-RSS sampler (Slice 1 + 2a)

justrach:
Partial fix for #124 (Slice 1 + 2a of the epic; #124 stays open for 2b).

## Problem

The root Agent runs on the process-init arena and **never resets it**, so every per-turn transient allocation accumulates for the life of a REPL / `--json` / serve session — unbounded RSS growth. `benchmarks/memory.py` only measures **one-shot** peak RSS, so it's structurally blind to this.

## Slice 1 — a session-RSS sampler that proves it

`benchmarks/memory_session.py` drives ONE persistent `graff --json` child through N turns and prints the RSS-vs-turn curve + a least-squares MB/turn slope. On unfixed `main`:

```
unfixed main:  +1.482 MB/turn   (20.5 -> 39.9 MB over 14 turns)   GROWING — leak present
```

## Slice 2a — a per-turn scratch arena (the transient-parse 80/20)

- Agent gains an **optional** `scratch_arena` (null on subagents/one-shots — they free their own arena per task, so `scratchAlloc()` falls back to `arena` and their behavior is unchanged). `main()` owns the storage and sets it on the **root only**.
- `request()` resets it (`.retain_capacity`) each call; all scratch data is consumed within a request.
- The high-volume transient parses move to scratch: the OpenAI per-event envelope parse (`assembleOpenAI`) and `isStreamEnd`'s bool-check parse.

**Correctness (the subtle part):** `assembleOpenAI` captured `role`/`id`/`name` as **slices into the parse tree** and stored them in the *persisted* assistant message — those are now **duped onto the session arena**, else the next request's reset is a use-after-free. The session benchmark **segfaulted on turn 2** before this dupe was added — caught and fixed. `assembleAnthropic` **must stay** on the session arena: it reuses the parsed `message`/`content_block` objects *as* the assistant message, so it can't move.

## Result

```
this change:   +0.922 MB/turn   (20.2 -> 33.1 MB over 14 turns)   — a 38% cut, no crash
```

`zig build test` **152/152**; the 14-turn session runs clean.

## Remaining (Slice 2b — #124 stays open)

The residual ~0.9 MB/turn is the per-request `normalizeOpenAIHistory`/`sanitizeMessagesUtf8` re-dupe, the mainloop steering intermediates, and history retention — which the spec deliberately deferred until a sampler existed to validate it. That sampler now exists, so 2b is unblocked.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/benchmarks/memory_session.py
+++ b/benchmarks/memory_session.py
@@ -1,0 +1,97 @@
+#!/usr/bin/env python3
+"""Session RSS-vs-turn: does a long-lived graff conversation leak memory?
+
+`memory.py` measures ONE-SHOT peak RSS, which is structurally blind to per-turn
+growth — a one-shot reads one line and exits before accumulation matters. This
+drives a SINGLE persistent `graff --json` child through N turns (stdin kept
+open), samples its RSS at each turn boundary, and prints the RSS-vs-turn curve
+plus a least-squares MB/turn slope over the post-warmup tail.
+
+Flat-after-warmup = fine. A positive slope is the #124 leak (the root Agent's
+session arena never resets, so per-SSE-event parse garbage accumulates for the
+life of the process) — now with a number to confirm or refute a fix.
+
+Run:  CODEGRAFF_API_KEY=... python3 benchmarks/memory_session.py [N] [graff-path]
+      N defaults to 16; graff-path defaults to `graff` on PATH.
+"""
+import subprocess, json, os, sys
+
+os.chdir(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+N = int(sys.argv[1]) if len(sys.argv) > 1 else 16
+GRAFF = sys.argv[2] if len(sys.argv) > 2 else "graff"
+
+# Short prompts that each stream a paragraph — many SSE events per turn (the
+# parse-garbage axis #124 is about) without slow multi-tool work, so the RSS
+# curve has plenty of samples. Swap in benchmarks/tasks.py for tool-heavy turns.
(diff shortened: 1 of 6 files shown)


justrach:
### Follow-up profiling: the residual is largely a **Debug-build allocator artifact**, not a release leak

Instrumented the session arena (`ArenaAllocator.queryCapacity` per turn) and profiled with `vmmap`:

- `session_arena_cap` is **flat at 1,719,400 bytes** across all turns, and `scratch_cap` is bounded (retains its high-water mark) — so **Slice 2a fully bounds the session arena** (its stated goal). History content (`msg_bytes`) grows only ~1 KB/turn (legit).
- The RSS growth I measured under `zig build` (**Debug**) is **`VM_ALLOCATE +22.88 MB`** over 6 turns while both arenas stay flat — the signature of Zig's **DebugAllocator retaining freed large allocations as guard pages** (use-after-free detection). Release builds don't do that.

So the "+0.92 MB/turn residual" was dominated by a **Debug-only measurement artifact**; the shipped **ReleaseFast** binary's session arena is bounded by this change. The "2b" work (normalize/steering deep-copy) is therefore lower priority than the PR implies — the arena is already flat.

Also worth noting: `gpt-5.6-sol` was asked to trace this independently and confidently blamed the `assembleOpenAI` arena accumulators — **contradicted by the measurement** (arena flat), a good reminder to verify against ground truth over a plausible code-read.

Issue #198, opened 2026-07-13
Interrupted sessions can leak codedb, xcodebuild, and XCTest/CoreSimulator process trees


yxlyx:
## Summary

Long-running/interrupted Codegraff terminal sessions can leave tool subprocesses and test process trees alive indefinitely. On one machine this accumulated enough processes/memory/load to make the laptop run hot, while the user's localhost Next.js server was essentially idle.

## Observed

Environment:

- `graff 0.0.201`
- `codedb 0.2.5829`
- macOS 26.4.1, arm64

Snapshot while diagnosing high heat/load:

- **15** `graff --yolo` processes
- **53** sleeping `codedb search ...` children
- **5** orphaned XCTest `launchd_sim` processes
- **699** CoreSimulator-related processes total
- roughly **1,700** processes system-wide
- 46 GB physical memory in use (with compression), and load averages initially around 9 / 30 / 28

Examples of `codedb search` children still alive:

```text
codedb search card                    elapsed 3-02:09:20
codedb search foregroundStyle(.green) elapsed 1-01:32:19
codedb search LanguageAwareTextField  elapsed 00:30:27
```

These were sleeping at 0% CPU but remained children of old sleeping `graff --yolo` sessions.

There was also an `xcodebuild test ...` process reparented to PID 1, plus three XCTest device trees with about 225 CoreSimulator processes each. `xcrun simctl list devices` showed no ordinary booted simulator, so these were XCTest clone/device processes rather than an intentionally booted simulator. The orphaned `xcodebuild` was idle but had not exited.

A detached `codedb <repo> cli-daemon` also showed intermittent CPU bursts (roughly 5–27% in repeated samples), but the larger lifecycle problem is the accumulation of stale CLI/tool/test processes.

For comparison, the Next.js process serving the page under investigation was normally at 0% CPU and only briefly used CPU while handling a request. Direct `graff` CPU was also low; the accumulated descendants and an active codedb daemon were the Codegraff-related load.

## Likely trigger

The machine had several long-running Codegraff sessions, including sessions that invoked `codedb search` and `xcodebuild test`. Some tool calls/sessions appear to have been interrupted, abandoned, or outlived their original shell. Exact minimal reproduction is not yet known.

## Expected

- Foreground tool subprocesses should be terminated and reaped when a tool call is cancelled/interrupted or its owning session exits.
- Process-group cleanup should cover grandchildren such as `xcodebuild` -> XCTest clone devices -> CoreSimulator services.
[...]

PR #200, opened 2026-07-13
Manage idle localhost servers and their process trees

### LINKED ISSUE #199: Auto-pause and stop Codegraff-started localhost servers after inactivity
## User problem

When Codegraff starts a localhost development/demo server, that server can remain alive for days after the user has stopped looking at it. It is easy to forget because it runs silently in an old terminal/session. Over time, multiple dev servers, file watchers, HMR workers, and related descendants can accumulate and consume memory, CPU, battery, ports, and thermal headroom without the user realizing it.

This was observed during the investigation in #198: a Next.js server on port 3002 had been alive for more than three days. It was mostly CPU-idle, but still retained memory, workers, file watchers, open browser connections, and a port. The broader machine had accumulated many other stale session-owned processes as well.

## Feature request

Codegraff should supervise localhost servers that **it starts** and apply an explicit idle lifecycle, for example:

1. **Running** while the server is actively being used.
2. **Paused/sleeping** after a configurable period with no meaningful user traffic or interaction.
3. **Stopped** after a longer configurable idle period.

The exact defaults and pause mechanism are open to design, but the default should prevent forgotten localhost servers from running indefinitely and invisibly.

## Important behavior

- Track the complete process group/tree, not only the initial `npm`/shell PID.
- Record the command, cwd, detected listening ports, owning Codegraff session, start time, and last meaningful activity.
- Do not treat HMR/WebSocket keepalives from a forgotten browser tab as meaningful user activity forever.
- Warn visibly before pausing/stopping and log what happened.
- Provide an easy resume/restart action using the original command.
- Allow users to **pin/keep alive** a server that intentionally needs to run indefinitely.
- Make pause and stop thresholds configurable globally and per server/project.
- On Codegraff/session exit, either stop the server tree or explicitly transfer/retain it with user confirmation.
- Only manage processes launched and owned by Codegraff. Never stop an unrelated service merely because it listens on localhost.

## UI/CLI visibility

There should be a place to see managed background resources, such as:

```text
localhost:3002  my-website  next dev  idle 42m  [Resume] [Stop] [Keep alive]
```

A CLI equivalent such as `graff processes`, `graff servers`, or `graff cleanup` would help users discover and clean up forgotten work.

## Design note: “pause”

A raw `SIGSTOP` would retain memory and the port and cannot itself detect a new request. Alternatives could include:

- suspend and require manual/UI resume;
- terminate the real server while a lightweight supervisor retains metadata and offers one-click restart; or
- keep a lightweight port proxy that restarts the server on the next request.

The implementation should make the state clear rather than claiming a process is paused while it still consumes substantial resources.

## Acceptance criteria

- [ ] A Codegraff-started localhost server is registered with its full process tree and listening port(s).
- [ ] After configurable inactivity, it transitions according to the idle policy and the user is notified.
- [ ] After the longer timeout, the full server process tree is gone and its resources/port are released.
- [ ] The original command can be restarted easily.
- [ ] A keep-alive/pin option disables idle cleanup for intentional long-running servers.
- [ ] Existing unrelated localhost services are never affected.
- [ ] Tests cover HMR child processes, stale browser WebSockets, session interruption, and process-tree cleanup.

Related lifecycle/process-leak report: #198.

yxlyx:
## Summary

- automatically move simple recognized localhost/dev-server commands (`npm run dev`, `pnpm preview`, `next start`, `vite`, `python -m http.server`, etc.) into supervised background jobs
- pause managed servers after 30 minutes of inactivity and stop them after 2 hours, with global environment overrides and a per-job `keep_alive` pin
- add `bash_resume` plus managed/paused/pinned/idle-stopped visibility in `/jobs`
- launch shell tools in their own POSIX process groups, so cancellation, timeout, `bash_kill`, idle stop, and session exit terminate npm/node/XCTest-style descendant trees rather than only `/bin/sh`
- document the lifecycle and configuration

Idle activity is deliberately local to the managed job: new stdout/stderr or an explicit `bash_output`/`bash_resume` call. Codegraff does not scan or interfere with unrelated localhost listeners.

Defaults can be configured with:

```sh
GRAFF_SERVER_PAUSE_MINUTES=15 GRAFF_SERVER_STOP_MINUTES=60 graff
GRAFF_SERVER_IDLE=off graff
```

## Tests

- `zig fmt --check` on all changed Zig files
- `zig build test` (includes command-recognition/lifecycle tests and a process-group cleanup integration test)
- `zig build -Doptimize=ReleaseFast`
- cross-build: `zig build -Dtarget=x86_64-linux -Doptimize=ReleaseFast`
- cross-build: `zig build -Dtarget=x86_64-windows -Doptimize=ReleaseFast`
- generated `graff --schema` parses as JSON and includes `bash_resume`

Closes #199.
Related to #198.

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,6 +419,7 @@ prefixes on body lines.
 /compact        summarize history into a fresh context
 /save | /resume | /sessions   session persistence; bare /resume → interactive picker
 /todo           show the current task list
+/jobs           list background jobs and managed localhost server state
 /mcp [add …]    list MCP servers/tools; /mcp add <name> <cmd> connects one live
 /help           list commands
 exit | /exit | ctrl-d | ctrl-c(empty)   quit
@@ -543,7 +544,8 @@ See [`sdk/README.md`](sdk/README.md).
 
 | Tool                 | Kind     | Implementation                                            |
 |----------------------|----------|-----------------------------------------------------------|
-| `bash`               | built-in | `std.process.run` → `/bin/sh -c`, stdout+stderr+exit code |
+| `bash`               | built-in | `/bin/sh -c`, capped output, managed background jobs      |
+| `bash_output`/`bash_resume`/`bash_kill` | built-in | inspect, resume, or stop background jobs and their process trees |
 | `read_file`          | built-in | `Io.Dir.cwd().readFileAlloc` (256 KB cap)                 |
 | `edit_file`          | built-in | exact string replace; unique match required unless `replace_all` |
 | `write_file`         | built-in | `Io.Dir.cwd().writeFile`                                  |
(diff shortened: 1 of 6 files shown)


justrach:
Closing as abandoned vs tip. Idle localhost supervision is parked (#199): jobs already cover tool-started bash; they do not pause forgotten `next dev` trees, and 282 is not starting a process-supervisor product. Reopen against tip if that product is wanted.

PR #207, opened 2026-07-14
fix(agent): context-overflow follow-up — recovery-path parity + local-model wedge (#201-#204)

### LINKED ISSUE #201: context-overflow: perOutputCap × keep_recent=4 lets recent tool outputs wedge the window (hole in #193 per-output cap)
## Symptom

The #193 follow-up (per-output cap + in-turn recovery, `b227ef8`) bounds any single
tool output before send and, on an over-window rejection, `emergencyTrim` /
`trimOldestToolOutputs` reclaims room and retries. But the trim keeps the
`keep_recent = 4` most-recent tool outputs **verbatim**, and `perOutputCap()` is
window-proportional at `context * 2` bytes (~50% of the window in estimated
tokens). So four recent large outputs can pin ~2x the window in content the
recovery cannot touch -> the retried request is still over-window -> hard wedge,
escapable only by a manual `/compact`.

This is a hole in what shipped in #193, not a new subsystem.

## Evidence

- `src/provider.zig:97-100` - `perOutputCap()` = `context * 2` bytes (~context/2 tokens per output)
- `src/agent_compact.zig:443,455` - `trimOldestToolOutputs` keeps 4 recent verbatim (`keep_recent = 4`)

## Invariant violated

`keep_recent * perOutputCap` must be <= what `emergencyTrim` can reclaim.
Currently `4 * ~50% window = ~200% window`, so recovery can never bring a
recent-output burst back under the wall.

## Proposed fix

Couple the two so the recent-output floor stays reclaimable:

- Lower `perOutputCap()` to ~1/8 of the window (`context / 2` bytes) so 4 recent
  outputs ~= 50% of the window, leaving headroom for the trimmed remainder +
  system prompt to fit on retry.
- Add an absolute ceiling (~256 KB, codex-style fixed cap) so huge-window models
  still bound one pathological single result.
- Optionally drop `keep_recent` 4 -> 2 for more margin.
- Sanity-check the constant against the smallest supported local window so the
  cap never exceeds the window itself.

## openai/codex comparison

`openai/codex` caps every exec/function output at an **absolute** ceiling
(default 10k tokens/bytes, `format_exec_output_str` -> `formatted_truncate_text`,
per-model `TruncationPolicy`) with **no** recency exemption. That fixed cap is why
one fat result rarely wedges it. We adapt (keep the window-proportional cap for
large windows, add the absolute ceiling).

## Refs

#193 (parent), #192

### LINKED ISSUE #202: context-overflow recovery hardening: usage char/4 fallback + visible truncation + meter re-estimate (anthropic/openai parity)
Three small, independent hardening items on the anthropic/openai context-overflow
recovery path. The codex/`.responses` path already does the mature version of each
(#174/#175/#195); this brings the other providers to parity and makes truncation
visible everywhere.

## G3 - `recordUsage` has no chars-per-token fallback (anthropic/openai)

When the API response omits `usage`, `recordUsage` returns early and the context
meter freezes at its stale value, so the between-turns compaction gate can never
fire -> contributes to the wedge.

- `src/agent_request.zig:475-476` - `const usage = root.get("usage") orelse return; if (usage != .object) return;`

The codex twin already handles this: `recordUsageResponses` (`:660-668`) floors
the meter at `req_body_len / 4` and `fullInputEstimateTokens`. openai/codex uses
`approx_token_count = (text.len()+3)/4` as a universal fallback.

**Fix:** thread `body.len` into `recordUsage` (both call sites at `:404`/`:467`
have `body` in scope) and, on missing/non-object usage, floor
`last_context_tokens` at `max(fullInputEstimateTokens(self), body.len/4)`. Success
path stays authoritative (these providers resend full history each turn, so
`usage.total` already covers system+tools - no #174 chained-undercount).

## G6 - truncation is silent (all providers)

`capOversizedToolOutputs` truncates in place but the call site discards the
reclaimed count, so nothing is surfaced to the user or the trace.

- `src/agent_request.zig:220` - `_ = self.capOversizedToolOutputs(...)`
- `src/agent_compact.zig:472-481`

**Fix:** capture the return and, when `> 0`, emit a trace note (`tracer.note`,
already used at `:378`) plus the user-facing status line. The model already sees
an inline `[tool output truncated ...]` marker. Optional polish: fold the original
byte size into the marker, codex-style (`Warning: truncated output (original token
count: N)`).

## G10 - every in-place trim zeroes the meter (all providers)

Both truncators set `last_context_tokens = 0`, opening a transient blind window
where the between-turns gate can't fire until the next response returns usage; a
successful overflow recovery even clobbers the window-pin back to 0.

- `src/agent_compact.zig:458` and `:479` - `if (reclaimed > 0) self.last_context_tokens = 0;`

**Fix:** set `last_context_tokens = self.fullInputEstimateTokens()` (the post-trim
estimate) instead of `0`, so the meter reflects the trimmed size and never blinds.
openai/codex never zeroes - its char/4 estimate is always a live floor.

## Refs

#193, #174, #195

### LINKED ISSUE #203: local/unknown models wedge on context overflow: 200k default window + system-prompt-blind estimate + substring-only detection
## Symptom

An unknown or locally-hosted model (LM Studio / mlx / any non-catalogued
`base_url`) can hit a hard, unrecoverable context wedge. The gaps chain:

1. **G8** - `contextFor()` falls back to `default_context = 200_000` for any model
   not in the baked/overlay tables, so the window is a guess and every downstream
   cap/threshold is mis-sized. `src/pricing.zig:303`, `src/provider.zig:150`.
2. **G4** - the pre-send gate uses `fullInputEstimateTokens()`, which serializes
   only `self.messages` and omits the system prompt + tool schemas, so it
   under-counts and the gate under-fires. `src/agent_request.zig:614-620`.
3. **G2** - when the backend rejects, `isContextOverflow()` matches ~6 English
   substrings with no structured `error.code` / HTTP-status check, so a
   local provider whose message differs never triggers recovery and the meter is
   never pinned. `src/agent_request.zig:81-92`.

Result: bad window guess -> under-count -> gate never fires -> backend rejects ->
needle miss -> no recovery -> hard wedge (only manual `/compact` escapes).

## Design note (from openai/codex research)

openai/codex has **no magic** for unknown local windows either - its fallback is a
single hardcoded 272k constant. Its real fix is escape hatches: a user config
override (`model_context_window`, clamped to max) plus a fetched/cached per-model
catalog (`list_models`). Note: OpenAI-compatible `/v1/models` for local servers
usually returns only `id`/`object`/`owned_by`, **not** a context window, so a probe
often falls back anyway - treat probing as a spike, not the plan.

## Proposed fix (priority order)

1. **Window override (root fix, do first):** a `--context` flag / config field that
   sets `Provider.context` directly, bypassing the 200k floor. Cascades to correct
   `compactAt` (G9), `perOutputCap` (G1), and the pre-send gate.
2. **G4 estimate baseline:** add the serialized system-prompt + tool-schema byte
   count (or a fixed baseline constant, cf. codex `BASELINE_TOKENS = 12000`) to
   `fullInputEstimateTokens` before the `/4`, so the lower bound fires before the
   wall.
3. **G2 structured detection (backstop):** in `isContextOverflow`, first inspect the
   parsed error object's `code` field and the HTTP status (400/413), then keep the
   substrings as a last-resort fallback. This is the one place we go **beyond**
   codex, which abandoned the non-streaming 400 path that is exactly the local
   wedge surface.
4. **(spike, optional)** opportunistic metadata probe when a local `base_url` model
   isn't in the table; accept only if a window field is actually present.

## Refs

#193, #192, #165

### LINKED ISSUE #204: context metering parity: same-format provider switch carries stale token count; compaction threshold hardcoded at 80%
Two low-severity parity items in the context-metering path.

## G5 - a same-format provider switch carries the old model's token count

`applyProviderInner` only resets `last_context_tokens` on a **cross-format** switch
(the `!same_format` branch clears it). A switch between two same-format providers
(e.g. two openai-shaped endpoints with different windows) keeps the previous
model's `last_context_tokens`, so the meter and the compaction gate are measured
against the new window with a stale number until the next response corrects it.

- `src/providers.zig:72-93` - reset only in the `!same_format` branch (`:81`)

**Fix:** also zero (or re-estimate) `last_context_tokens` in the same-format path,
letting the next `recordUsage` / `fullInputEstimateTokens` refill it. openai/codex
re-resolves `ModelInfo` per model slug so identity always drives the number.

## G9 - compaction threshold hardcoded at 80% with no override or soft warning

`compactAt()` is `context / 10 * 8` for every provider with no config/env override,
and there is no "approaching limit" cue beyond the live context counter.

- `src/provider.zig:85-87`

**Fix:** thread an optional override into `compactAt()` from config/env; keep the
existing live `compact@Nk` / context display visible as the soft warning. Do **not**
copy codex's one-directional clamp (it hard-clamps the user value to <=90%, an
active UX complaint) - allow lowering *and* raising.

## Refs

#193

justrach:
Follow-up to #196, closing the ranked backlog from the context-overflow audit.
Together these break the compound local-model wedge chain end to end: bad window
guess -> undercount -> gate never fires -> needle miss -> no recovery -> hard wedge
(previously escapable only by a manual `/compact`).

**Stacked on #196** (base = `fix/193-tool-output-cap`); GitHub will retarget this to
`main` when #196 merges, at which point the `Closes` keywords fire.

## Commits
- `ea230cc` — G1/G3/G6/G10 (#201, #202): the recovery-path holes
- `3a603d0` — G8/G4/G2/G5/G9 (#203, #204): local-model window + metering parity

## What changed
| Gap | Fix |
|---|---|
| **G1** #201 | `perOutputCap` was ~50% of the window each; with `keep_recent=4` verbatim, four recent outputs pinned ~2x the window past recovery. Now ~1/8 window + 256 KB ceiling, so 4 recent stay reclaimable. |
| **G3** #202 | `recordUsage` froze the meter on missing usage; now floors at `max(fullInputEstimateTokens, body/4)` (the fallback the codex path already had). |
| **G6** #202 | `capOversizedToolOutputs` truncated silently; now emits a tracer note + status line. |
| **G10** #202 | in-place trims zeroed the meter; now re-estimate the trimmed size. |
| **G8** #203 | unknown/local models defaulted to a 200k guess. `GRAFF_CONTEXT` declares the real window, applied only to non-catalogued models (`pricing.isKnownModel`). |
| **G4** #203 | the pre-send gate estimate omitted system+tools; adds a window-clamped baseline (kept out of `fullInputEstimateTokens`, which stays pure). |
| **G2** #203 | overflow detected only by English substrings; now matches structured `error.code` first (codex parity), substrings as fallback. |
| **G5** #204 | a same-format provider switch kept the old model's token count; now re-estimates. |
| **G9** #204 | 80% compaction threshold hardcoded; `GRAFF_COMPACT_PCT` overrides it (both directions). |

Design note: the openai/codex research behind the local-model decisions (window
override vs probe vs catalog, structured detection) is captured in #203.

## Tests
`zig build test` 183/183 (was 179; +perOutputCap, +recordUsage-fallback,
+contextWindowFor, +compactAt-override; overflow-detection tests gain structured-code
cases), `zig fmt` clean.

Closes #201
Closes #202
Closes #203
Closes #204

### DIFF
--- a/scripts/test-pty-overflow.py
+++ b/scripts/test-pty-overflow.py
@@ -1,0 +1,186 @@
+#!/usr/bin/env python3
+"""End-to-end real-PTY test for anthropic/openai context-overflow handling (#201-#203).
+
+Points graff at a local OpenAI-compatible backend (the built-in `lmstudio` provider,
+http://127.0.0.1:1234) whose only reply is an injected error, and drives a real turn
+through the terminal. Two scenarios prove the overflow detection is BOTH correct and
+precise:
+
+  A. error.code = context_length_exceeded with a message that matches none of the
+     English substrings (Dutch). graff must still detect the overflow via the
+     STRUCTURED code (#203/G2), pin the meter to the window, and stay responsive
+     rather than wedge (#201/#202). Observable: the prompt's ctx meter reads
+     "<W>k/<W>k ctx (100% ...)".
+  B. error.code = rate_limit_exceeded with a non-overflow message. graff must NOT
+     mistake it for an overflow: the meter must not pin. Proves detection is precise.
+
+Requires 127.0.0.1:1234 to be free (the lmstudio provider URL is fixed); skips if a
+real LM Studio (or anything) already holds it.
+"""
+
+import http.server
+import json
+import os
+import re
+import socket
+import sys
+import tempfile
+import threading
+
+from pty_harness import PtySession, terminal_text
+
+
+_arg = sys.argv[1] if len(sys.argv) > 1 else "graff"
+GRAFF = os.path.abspath(_arg) if os.sep in _arg else _arg
+
+# "<used>k/<window>k ctx (<pct>% · compact@<X>k)" — the "·" is U+00B7.
+METER_RE = re.compile(r"(\d+)k/(\d+)k ctx \((\d+)% · compact@(\d+)k\)")
+PINNED_RE = re.compile(r"(\d+)k/(\d+)k ctx \(100% · compact@\d+k\)")
+
+
+class OpenAiErrorMock:
+    """Serves one fixed OpenAI-style error envelope for every /v1/chat/completions."""
+
+    def __init__(self, error_obj: dict) -> None:
+        self.body = json.dumps({"error": error_obj}).encode()
+        self.hits = 0
+        parent = self
+
+        class Handler(http.server.BaseHTTPRequestHandler):
+            def do_POST(self) -> None:  # noqa: N802
+                parent.hits += 1
+                length = int(self.headers.get("content-length", 0))
+                if length:
+                    self.rfile.read(length)
+                self.send_response(200)
+                self.send_header("content-type", "application/json")
+                self.send_header("content-length", str(len(parent.body)))
+                self.end_headers()
+                self.wfile.write(parent.body)
+
+            def do_GET(self) -> None:  # noqa: N802 (e.g. a /v1/models probe)
+                self.send_response(404)
+                self.end_headers()
+
+            def log_message(self, *_a) -> None:  # silence the default stderr logging
+                pass
+
+        self.httpd = http.server.ThreadingHTTPServer(("127.0.0.1", 1234), Handler)
+
+    def start(self) -> None:
+        threading.Thread(target=self.httpd.serve_forever, daemon=True).start()
+
+    def stop(self) -> None:
+        self.httpd.shutdown()
+        self.httpd.server_close()
+
+
+def _run(error_obj: dict, tmp: str):
+    """Run one turn against a mock returning error_obj; return (rendered_text, hits)."""
+    mock = OpenAiErrorMock(error_obj)
+    mock.start()
+    try:
+        env = {
+            "HOME": tmp,
+            "LMSTUDIO_API_KEY": "local-pty-test",
+            "GRAFF_FLEET": "off",
+            "GRAFF_NO_TELEMETRY": "1",
+        }
+        ambient = tuple(
+            k for k in os.environ
+            if (k.startswith("GRAFF_") or k.startswith("CODEX_") or k == "NO_COLOR")
+            and k not in env
+        )
+        with PtySession(
+            GRAFF,
+            ["--model", "lmstudio", "--no-telemetry"],
+            cwd=tmp,
+            env=env,
+            unset_env=ambient,
+            timeout=20.0,
+        ) as session:
+            session.wait_for_literal("] ›")
+            cursor = len(session.raw)
+            session.send_line("hello")
+            # The turn ends with an api error either way; wait for it, then settle.
+            session.wait_for_literal("api error:", start=cursor)
+            session.pump_for(1.5)
+            rendered = terminal_text(bytes(session.raw[cursor:]))
+
+            # Session must remain usable after the failed turn (no wedge): a local
+            # command still works and the REPL exits cleanly.
+            c2 = len(session.raw)
+            session.send_line("/help")
+            session.wait_for_literal("/models [health]", start=c2)
+            session.send_key("ctrl-d")
+            result = session.read_until_exit(5.0)
+            if result.timed_out or result.exit_code != 0:
+                raise SystemExit(
+                    f"REPL did not exit cleanly: exit={result.exit_code} "
+                    f"timed_out={result.timed_out}"
+                )
+            return rendered, mock.hits
+    finally:
+        mock.stop()
+
+
+def main() -> None:
+    # The lmstudio provider URL is hardcoded to :1234; bail cleanly if it's taken.
+    probe = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    # SO_REUSEADDR binds over a TIME_WAIT port left by a prior run, but still fails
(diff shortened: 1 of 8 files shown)


justrach:
Added an **end-to-end overflow PTY test** (`scripts/test-pty-overflow.py`) that points graff at a mock `lmstudio` backend injecting a `context_length_exceeded` error and drives a real terminal turn.

Writing it **caught a real gap in the G2 fix**: the structured-`error.code` detection only covered the anthropic-style `{"type":"error"}` envelopes. OpenAI-compatible errors are `{"error":{...}}` (no top-level `"type"`) and land on the generic `apiErrorMessage` path, which was still passing `null` for the code — so a local provider (LM Studio / mlx / any openai-compat `base_url`) whose message text we don't match on would **not** have been recognized as an overflow. That's the exact case G2 exists for. Fixed with `errorCode(root)` (`aa2b995`).

The test proves both directions:
- `code=context_length_exceeded` + a **Dutch** message (matches no English needle) → meter pins to `200k/200k (100%)` and the REPL stays responsive → detection went through the structured code, no wedge.
- `code=rate_limit_exceeded` + non-overflow message → meter must **not** pin → detection is precise.

Deterministic across repeated runs (`SO_REUSEADDR` on the port probe); skips cleanly if `127.0.0.1:1234` is already held by a real LM Studio.

Issue #142, opened 2026-07-10
REPL table renderer: dispWidth counts codepoints, not display width (CJK/emoji misalign borders)


justrach:
Follow-up from PR #127 (merged into release/v0.0.193, commit 16c9db7).

`dispWidth` in `src/repl.zig` advances by `utf8ByteSequenceLength` and counts every codepoint as width 1. East-Asian full-width characters (width 2) and emoji make the box-drawing borders (`│ ┌ ┼ …`) misalign in the TUI table renderer.

**Fix:** use an East-Asian-width table (wcwidth-style) so `dispWidth` returns 2 for wide codepoints. Affects column-width computation and cell padding in `renderTable`/`renderRecords`/`wrapCell`.

The GUI path is unaffected (browser handles glyph width). Non-blocking; TUI-only cosmetic misalignment for CJK/emoji content.

PR #211, opened 2026-07-14
fix(agent): width-aware interactive prompt/status line (#209)

### LINKED ISSUE #209: CLI prompt/status line is cut off or splits badges in narrow terminal panes
## Summary

The interactive CLI prompt/status line is not width-aware. In a narrow terminal pane it wraps or appears cut off in the middle of a badge; for example, the visible line can end at:

```text
[gpt-5.6-sol · High · co█
```

This makes the cursor appear to be inside the `codex` provider badge and makes it unclear where user input begins.

## Environment

- graff `0.0.201`
- macOS / Ghostty (`TERM=xterm-ghostty`)
- Codex model `gpt-5.6-sol`

## Reproduction

Start graff in a narrow terminal pane, or reproduce under tmux with a 28-column pane:

```sh
tmux new-session -d -x 28 -y 12 -s graff-narrow \
  "graff --model gpt-5.6-sol --no-telemetry"
tmux capture-pane -p -t graff-narrow
```

The status prompt physically wraps like this:

```text
[gpt-5.6-sol · High · codex
· cwd /path/to/project
] ›
```

Depending on the pane/viewport, only the first partial row is visible, which looks like a truncated prompt with the cursor embedded in the provider label.

## Expected

The prompt should adapt to available width. Possible behavior:

- omit lower-priority metadata (`cwd`, token/cache/cost detail) at narrower widths;
- never split a badge such as `codex` across the edge;
- and/or put a short input marker on a clean new row.

The cursor/input location should remain visually unambiguous.

## Likely cause

`Agent.prompt` in `src/agent.zig` emits the full model, effort, provider, modes, cwd, and optional context/cache/cost string without budgeting against terminal width. `readline.zig` checks the cursor column afterward and may move input to another row, but by then the status string itself has already wrapped.

A narrow-width PTY regression covering the main `Agent.prompt` (not only the model/settings pickers) would help.

justrach:
## Problem

`Agent.prompt` emitted the full model/effort/provider/mode/cwd/context status line with no width budget, so a narrow terminal pane soft-wrapped it mid-badge (e.g. splitting `codex`) and left the cursor visually inside a label:

```
[gpt-5.6-sol · High · co
```

`readline.zig` only reacts afterward by moving the *input* to a fresh row; by then the status string has already wrapped.

## Fix

`prompt()` budgets each segment against `termCols()`:

- The model plus the settings that disambiguate the cursor (effort, provider, mode badges) are offered first; `cwd`/context/cache/cost are dropped when they no longer fit.
- The context meter outranks `cwd` for the budget (it is the urgent, changing signal near the compaction threshold) but still renders after `cwd`, so the familiar left-to-right order is kept when both fit.
- The line reserves the fixed frame (`[` … `] › `) plus a column of slack, so nothing but a model name longer than the pane can wrap; `readline` then gives the input its own row as before.

Progressive disclosure at increasing widths (captured from a real PTY):

```
cols=28   [deepseek-v4-pro] ›
cols=34   [deepseek-v4-pro · Medium] ›
cols=50   [deepseek-v4-pro · Medium · codegraff] ›
cols=120  [deepseek-v4-pro · Medium · codegraff · cwd /…] ›
```

The wcwidth-style `dispWidth`/`codepointWidth` table moves from `repl.zig` into `term.zig` (the shared terminal-utilities leaf) so the TUI table math and this budgeting share one definition; `repl.zig` delegates and keeps its #142 test.

## Tests

- New `scripts/test-pty-narrow-prompt.py` — drives a real 28-col PTY and asserts the line fits, badges stay whole, and `cwd` is dropped, while a wide pane still shows `cwd`.
- `scripts/test-pty-repl.py` widened to 200 cols so its badge-state session fits the full line.
- `zig build test` 184/184, `zig fmt` clean, full PTY suite green.

Closes #209

### DIFF
--- a/scripts/test-pty-narrow-prompt.py
+++ b/scripts/test-pty-narrow-prompt.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python3
+"""Real-PTY test for the width-aware interactive prompt/status line (#209).
+
+Before the fix, `Agent.prompt` emitted the full model/effort/provider/cwd/context
+status line with no width budget, so a narrow terminal pane soft-wrapped it mid-
+badge (e.g. splitting `codex`) and left the cursor visually inside a label. The
+fix budgets each segment against `termCols()`: the model and the badges that
+disambiguate the cursor come first, and cwd/context/cache/cost are dropped when
+they no longer fit, so the rendered line never exceeds the pane width and no
+badge is split at the edge.
+
+This drives a real terminal at several widths and asserts:
+  * narrow pane  -> the prompt line fits the pane, its badges stay whole, and
+    low-priority metadata (cwd) is dropped;
+  * wide pane    -> the same launch shows the full metadata (cwd), proving the
+    narrow layout is genuine width budgeting and not an unconditional strip.
+
+Rendered (not raw) so the check is independent of the terminal's newline
+translation. `-`/`>`-style separators are single display columns and the cwd is
+dropped in the narrow case, so len() of the rendered line equals its display
+width here.
+"""
+
+import os
+import sys
+import tempfile
+
+from pty_harness import PtySession, terminal_text
+
+_arg = sys.argv[1] if len(sys.argv) > 1 else "graff"
+GRAFF = os.path.abspath(_arg) if os.sep in _arg else _arg
+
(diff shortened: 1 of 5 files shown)


justrach:
Already on main (width-aware prompt render). Closing as superseded.

PR #212, opened 2026-07-14
feat(gui): render LaTeX math in chat via KaTeX (#208)

### LINKED ISSUE #208: GUI chat renders LaTeX equations as raw source
## Summary

The GUI chat Markdown renderer displays LaTeX source and delimiters literally instead of rendering mathematical notation. This is especially visible in technical/system-design answers where the agent emits display math using `\[ ... \]`.

Example rendered output:

```text
\[
G \approx 1+\alpha+\alpha^2+\cdots+\alpha^m.
\]
```

The user sees the backslashes, commands, and delimiters rather than a formatted equation.

## Reproduction

Have an assistant response contain:

```md
\[
G \approx 1+\alpha+\alpha^2+\cdots+\alpha^m.
\]
```

The parser behavior can also be reproduced directly:

```sh
cd gui
bun -e 'import { parseMarkdown } from "./src/components/chat/markdown/parser.ts"; console.log(JSON.stringify(parseMarkdown(String.raw`\[
G \approx 1+\alpha+\alpha^2+\cdots+\alpha^m.
\]`), null, 2));'
```

Current result is one paragraph/text node whose literal value contains the complete LaTeX source.

## Expected

Common agent-emitted math delimiters render as formatted math, at minimum:

- display: `\[ ... \]` and `$$ ... $$`
- inline: `\( ... \)` and `$ ... $`

During streaming, an unclosed delimiter should degrade safely without crashing or swallowing the rest of the response.

## Actual

The raw delimiters and LaTeX commands are displayed verbatim.

## Cause

The GUI uses a custom dependency-free parser. `gui/src/components/chat/markdown/types.ts` has no inline or block math AST node, and `parser.ts` only recognizes headings, paragraphs, fenced code, blockquotes, lists, tables, thematic breaks, emphasis, inline code, and links. `MarkdownRenderer.tsx` consequently has no math rendering branch. `gui/package.json` also has no KaTeX/MathJax dependency.

The comment in `types.ts` says the AST covers “the constructs the chat agent actually emits,” but agent responses do emit LaTeX equations.

## Suggested coverage

Add parser and renderer tests for:

- `\[x^2\]`
- `$$x^2$$`
- `before \(x^2\) after`
- `before $x^2$ after`
- escaped dollar signs
- inline/code fences containing dollar signs or LaTeX, which must remain code
- incomplete delimiters during streamed responses
- unsafe HTML-like LaTeX input remains sanitized

justrach:
## Problem

Agent responses emit LaTeX, but the dependency-free chat Markdown parser had no math node. Display formulas and inline math rendered as **raw source**:

```
\[
G \approx 1+\alpha+\alpha^2+\cdots+\alpha^m.
\]
```

The user saw backslashes, commands, and delimiters instead of typeset notation.

## Fix

**Parser** (`types.ts`, `parser.ts`) — adds an inline and a display `math` AST node and recognizes all four agent-emitted forms:

| form | example | level |
| --- | --- | --- |
| `\[...\]` | multi-line display | block |
| `$$...$$` | display | block |
| `\(...\)` | inline | inline |
| `$...$` | inline | inline |

Guardrails, all covered by tests:
- Inline **code keeps top precedence**, so a `$` inside backticks stays code; **fences are captured before math**, so `$$` inside a fenced block stays code.
- The `$...$` form is **deliberately narrow** (opener/closer hug the content, never against a digit, never after a backslash), so prose currency like `$5 and $10` and an escaped `\$` are left as text.
- An **unclosed delimiter** (still streaming) degrades to open math / literal text instead of throwing or swallowing the rest of the message.

**Renderer** (`KatexMath.tsx`, `MarkdownRenderer.tsx`) — a new component typesets the captured LaTeX with **KaTeX**:
- `renderToString` is pure (no DOM), so it stays server-renderable and unit-testable.
- `throwOnError: false` → malformed/streaming input renders as a visible KaTeX error rather than throwing.
- `trust: false` → `\href`/raw-HTML commands can't inject anything unsafe (verified: no anchor, no `javascript:` href, raw `` escaped).
- `katex.min.css` is bundled via the existing `@import` block in `styles/index.css` (mirroring `shadcn/tailwind.css`), so the component carries no CSS import.

## Tests

- `parser.test.ts` **+7**: display/inline parsing, code + fence protection, escaped `$`, prose currency, streaming degradation.
- `MarkdownRenderer.test.tsx` **+4** (full pipeline via `renderToStaticMarkup`): display + inline typeset to KaTeX, `$` in code stays code, unsafe `\href` emits no anchor / `javascript:` href.
- gui `bun test`: **146/146**; eslint clean; new files typecheck clean under strict `tsc`.

## Honest note on verification

The full Tauri/Vite build (`bun run build`) is **not** exercised in my environment (its `gen:desktop-contracts` step needs cargo), so the bundled-CSS **visual** result wasn't run here. The HTML pipeline is fully unit-tested and the CSS `@import` follows the repo's existing pattern; a quick visual sanity check on a real build is worthwhile before merge.

## Dependency

Adds `katex` (~0.17) + `@types/katex`. Follows the existing library-backed-render precedent (`MermaidDiagram`).

Closes #208


justrach:
Already on main (gui KaTeX renderer + #208 tests). Closing as superseded.

PR #213, opened 2026-07-14
fix(agent): pair the Codex WS delta boundary with a fresh response id (#194)

### LINKED ISSUE #194: Codex WS can pair a stale response ID with an advanced delta boundary
`stepResponses()` currently advances `codex_sent_upto` whenever a WS is live, but only replaces `codex_prev_id` when the parsed response contains a new `response.id`.

If `parseResponses()` returns `.ok` from output items without a terminal response ID (it accepts `items.items.len > 0`), the old ID remains while the boundary advances. The next delta request references the old server response but slices local input after the new boundary, omitting state the referenced response never held.

Relevant paths:
- `src/agent_request.zig`: `parseResponses()` accepts output items without `response.completed`/ID.
- `src/agent_steps.zig`: `codex_sent_upto` advances before/without requiring a new ID.
- `src/agent_request.zig`: `buildBody()` then pairs stale `previous_response_id` with the advanced slice.

Safe behavior: only advance the delta boundary when a new response ID is installed atomically; otherwise close/reset the WS continuation state so the next request fully re-anchors.

justrach:
## Problem

`stepResponses()` advances the Codex WS delta boundary (`codex_sent_upto`) whenever a WS session is live, but only replaces `codex_prev_id` when the parsed response contains a new `response.id`. `parseResponses()` accepts output items **without** a terminal response id, so a response can lack one — leaving the **old** id paired with an **advanced** boundary.

`buildBody()` then sends `previous_response_id: <old id>` + `messages[codex_sent_upto..]`, slicing off history the referenced response never held.

## Fix

Move the id and the boundary in lockstep (`agent_steps.zig`):
- New valid id → install it and advance `codex_sent_upto` together.
- No usable id (or a failed dup) → reset both to `null` / `0` so the next request re-anchors with the full input (`buildBody`'s `prev_id == null` path), instead of replaying a stale delta.

No WS teardown is needed — resetting the watermarks is enough; the next request sends the full input over the live session, exactly like the first request of any WS session.

## Test

A unit test drives a codex-WS agent through `stepResponses()`:
- **id-less response** → `codex_prev_id` is dropped and the boundary resets to 0 (full re-anchor).
- **id-bearing response** → the id installs and the boundary advances in lockstep.

Leak-checked (testing allocator). `zig build test` green, `zig fmt` clean, and `scripts/test-pty-codex-ws.py` (WS-fail → SSE fallback, WS-off) still passes.

Closes #194

### DIFF
--- a/src/agent_compact.zig
+++ b/src/agent_compact.zig
@@ -651,6 +651,72 @@ test "buildBody (.responses): delta while WS live; full input after closeCodexWs
     try std.testing.expect(std.mem.indexOf(u8, rebuilt_body, "third — not yet sent") != null);
 }
 
+// (#194) stepResponses must move the Codex WS delta boundary (codex_sent_upto)
+// and previous_response_id together. parseResponses accepts a response whose
+// output items carry no terminal response.id; advancing the boundary while
+// keeping the previous id would pair a stale id with a slice that id never held.
+// With a fresh id both advance in lockstep; with no id both reset so the next
+// request re-anchors with the full input.
+test "stepResponses (codex-ws): delta boundary only advances with a fresh response id (#194)" {
+    var arena_state = std.heap.ArenaAllocator.init(std.testing.allocator);
+    defer arena_state.deinit();
+    const a = arena_state.allocator();
+
+    var dummy_ws: ws.WsClient = undefined; // stepResponses only checks codex_ws != null
+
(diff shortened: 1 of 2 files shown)


justrach:
Already on main; Codex WS error handling is now #693 on #696. Closing as superseded.

Issue #193, opened 2026-07-12
codex/gpt-5.x: a single-turn tool-output burst overflows the context window before auto-compaction fires (no pre-send local-estimate gate)


justrach:
## Symptom

On a `codex` session (`gpt-5.6-sol`), a turn that fired several large tool calls at once (6 parallel `read_file` of big TSX/page files) died with:

```
codex api error: Your input exceeds the context window of this model. Please adjust your input and try again.
```

The turn is lost with a visible hard error, mid-work.

## Root cause: no pre-send local-estimate gate

Auto-compaction gates on `last_context_tokens`, which is only set from the **server's returned usage** (`recordUsage` / `recordUsageResponses` in `agent_request.zig`). That number **lags the locally-appended tool results**: within a single turn we append N large tool outputs to `self.messages` and then call `request()` again — but `last_context_tokens` still reflects the *previous* response's usage, from before the reads landed.

The 80% compaction gate (`mainloop.zig:654`, `provider.compactAt()`) is only checked **between turns**, not before each model request inside the agentic loop. So a burst of tool output in one turn is serialized and sent un-compacted, and the backend rejects it at the ~272k input wall.

We already have the right primitive but don't use it here: `fullInputEstimateTokens()` (`agent_request.zig:487`, byte/4 heuristic over the full history) exists, but is only consulted **after** a response, to correct the meter upward (`recordUsageResponses:533`). It is never used as a **pre-send** gate.

`contextFor()` correctly clamps every codex path to `codex_context_window` (270k) via `@min`, so `compactAt()` = 216k is right — the window number is not the bug. The bug is purely *timing*: we measure too late.

## What already partially covers this
- **#174** — on the `exceeds the context window` error, pins the meter to the window so the next turn compacts. Recovers, but the current turn still fails.
- **#163** — compaction / emergency-trim recover an over-cap history that fails to WRITE.

Both are *recovery after the fact*. We want *prevention*.

## What codex and opencode do (prevention)

**openai/codex** (`codex-rs/core`)
- Local `estimate_token_count` (byte heuristic) in `context_manager/history.rs`.
- `run_pre_sampling_compact` at the **start of every turn**, plus **mid-turn** `run_auto_compact(CompactionPhase::MidTurn)` when `token_limit_reached` — i.e. the overflow check runs before *every* model call, not just between user turns.
- Caps individual tool outputs: `truncate_function_output_payload` + `TruncationPolicy`; `trim_function_call_history_to_fit_context_window`.
[...]

PR #214, opened 2026-07-14
test(agent): lock in aggregate gating of a parallel tool batch (#192)

### LINKED ISSUE #192: Codex can still overflow between a parallel tool batch and its continuation request
## Symptom

A Codex turn requested six `read_file` calls in parallel and the immediate continuation failed with:

```
codex api error: Your input exceeds the context window of this model. Please adjust your input and try again.
```

The model was `gpt-5.6-sol` via the Codex provider.

## Why the previous fixes do not fully cover it

- #88 recovers an errored turn only when `last_context_tokens >= compactAt()`.
- #163 makes compaction/emergency trim survive old tool-output churn, but that logic runs only once compaction is invoked.
- #174 fixes the stale cross-turn WS/full-resend meter and recognizes the raw rejection, but it does not add an in-turn pre-sampling compaction gate.
- Responses tool outputs are individually capped at 16,384 chars, but a parallel batch is not checked as an aggregate before the next model request.

`Agent.runTurn()` currently does `request -> stepResponses/runTools -> request` with no context check between the tool results being appended and the continuation request. Normal auto-compaction is in `mainloop.zig` after a completed turn, which is too late for this path.

## OpenAI Codex comparison

The upstream `openai/codex` turn loop checks active context after sampling/tool execution and, when another sampling iteration is needed, runs `MidTurn` compaction before the continuation request (`codex-rs/core/src/session/turn.rs`). It also combines server usage with local estimates for newly appended items.

## Proposed fix

1. Port/retain #174's full-history estimate and context-rejection recovery on the affected release branch.
2. Add a pre-request context gate inside `Agent.runTurn()`, including between tool-loop iterations.
3. Refresh the Responses estimate after normalizing newly appended tool outputs, so a parallel batch can trigger compaction before it is sent.
4. Add a regression test with several parallel-sized `function_call_output` items crossing `compactAt()`.

justrach:
## TL;DR

**No code change is needed** — #192's overflow window is already closed on `main` by the #193 pre-send gate. This PR adds the regression test the issue asked for (its ask #4) and documents why the gate covers the parallel-batch case.

## Why it's already handled

`Agent.runTurn()`'s loop is:

```
gate (#193 inputOverCompactThreshold)  ← top of loop
  → request()
  → stepResponses()   ← appends the whole parallel tool batch to self.messages
  → loop back → gate  ← re-estimates over the appended batch
  → continuation request()
```

The #193 gate sits **exactly** between the tool batch being appended and the continuation request, and `inputOverCompactThreshold` → `fullInputEstimateTokens` serializes the **full** `self.messages` (including the freshly appended outputs). So a parallel batch that crosses `compactAt()` only in aggregate triggers compaction **before** the continuation is sent — which is #192's scenario.

(The #203 prefill baseline, in the in-flight `.202` follow-ups, only makes the gate fire earlier / more conservatively; it isn't required to close this window.)

## Test (the issue's ask #4)

`inputOverCompactThreshold (#192): a parallel tool batch is gated as an aggregate`:
- one `function_call_output` (~6 KB, individually far under the window and the 16 KB per-output cap) → gate does **not** fire;
- a six-output batch (the reported six parallel `read_file` calls) → aggregate serialized input crosses `compactAt()` → gate **fires**.

Verified on `main`: `zig build test` green, `zig fmt` clean.

Closes #192

### DIFF
--- a/src/agent_request.zig
+++ b/src/agent_request.zig
@@ -542,6 +542,39 @@ test "inputOverCompactThreshold (#193): local estimate gates a pre-send compact"
     try std.testing.expect(!inputOverCompactThreshold(&agent));
 }
 
+// (#192) A parallel tool batch must be gated as an AGGREGATE. runTurn's #193
+// pre-send gate sits at the top of the loop — after stepResponses appends the
+// whole batch and before the continuation request (agent.zig) — so
+// inputOverCompactThreshold re-estimates over every freshly appended output.
+// Six parallel read_file outputs, each individually well under the window (and
+// the 16 KB per-output cap), still trip the gate together, so compaction runs
+// before the continuation instead of overflowing the model.
+test "inputOverCompactThreshold (#192): a parallel tool batch is gated as an aggregate" {
+    var arena_state = std.heap.ArenaAllocator.init(std.testing.allocator);
+    defer arena_state.deinit();
+    const a = arena_state.allocator();
+    var msgs = std.json.Array.init(a);
+    var agent: Agent = undefined;
+    // compactAt() = 10_000/10*8 = 8_000 tokens ≈ 32_000 serialized bytes.
+    agent.provider = .{ .id = "codex", .kind = .responses, .auth = .bearer, .url = "", .api_key = "", .model = "gpt-5", .context = 10_000 };
+
+    // One ~6 KB output — individually far under the window — must not trip it.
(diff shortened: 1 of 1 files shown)


justrach:
Already on main under a later rewrite. Closing as superseded.

Issue #134, opened 2026-07-09
Stream stall incorrectly reported as user Esc interruption


yxlyx:
## Summary

During a Codegraff session, the assistant turn ended as `interrupted (esc)` even though the user says they did not press Esc. The UI also showed `stream stalled — ending turn` immediately before the interruption.

This caused the assistant response to be recorded as `[response interrupted by user]` despite no user interruption.

## What happened

Sequence observed in the chat UI screenshot:

1. Assistant ran a `bash` inspection command successfully.
2. Assistant ran `edit_file` successfully:
   - `replaced 1 occurrence(s) in lawbook/src/components/AskAgent.tsx`
3. The UI showed `thinking…`
4. Then the UI showed:
   - `stream stalled — ending turn`
   - `interrupted (esc)`
5. The next visible assistant message was `[response interrupted by user]`.

The user explicitly reported: “it self interrupted i didnt click esc”.

## Trace observations

I inspected `harness.trace.jsonl` in the working directory after the incident.

Relevant observations:

- The trace file is not clean text JSONL; it contains NUL bytes. `read_file` reported it as binary.
- Searching the trace for `interrupted`, `stream stalled`, `ending turn`, or `esc` found no matches.
- The last events before the user's report were normal API/tool events, e.g.:
  - `edit_file` succeeded around trace event 590: `{"name":"edit_file","result_bytes":63,"is_error":false}`
  - subsequent main API events were logged as `is_error:false`
- There does not appear to be a trace event recording the UI-side stall/interruption reason.

## Expected behavior

If the user did not press Esc, the turn should not be marked as user-interrupted. If the stream stalls and the harness decides to end the turn, it should be surfaced as a stream/harness error or retryable stall, not as `interrupted (esc)`.

Also, the event trace should record the stall/interruption cause so this can be debugged after the fact.

## Actual behavior

The UI marked the turn as `interrupted (esc)` and the conversation received `[response interrupted by user]`, even though no Esc key was pressed.

## Environment/context

- Model/session shown in trace: `gpt-5.5`
- Working directory: `/Users/limyuxi/lawbook`
- Harness trace file: `harness.trace.jsonl`
- The trace had many normal `api`/`tool` events and no explicit interrupt/stall event.

## Possible fixes / asks

- Distinguish actual keyboard/user cancellation from automatic stream-stall termination.
- Log a structured trace event for `stream stalled — ending turn`, including timeout duration and reason.
[...]

Issue #56, opened 2026-06-20
Harden SSE stall watchdog: guard peekByte + receiveHead, make stalls retryable (reconnect)


justrach:
## Summary

graff's streaming reader (`postStream`) could hang **forever at "thinking..."** when a connection went half-open mid-stream (the `HttpConnectionClosing` / `WriteFailed` / truncated-TLS case): the retry re-entered `postStream`, the spinner restarted, then the resumed read wedged - "reconnect doesn't reconnect."

## Root cause (confirmed against the code)

The 2026-06-17 idle-stall watchdog (racing `streamStallTask` against the line read via `Io.Select`) guarded the **SSE line read** but left sibling blocking reads unguarded, and surfaced stalls as a non-retryable error:

1. **`reader.peekByte()`** (the post-line "is there more?" probe, ~`main.zig:8638`) ran *after* the watchdog was cancelled. On a half-open socket (server gone, no FIN/RST) it blocks forever. *(Primary hang.)*
2. **Watchdog stalls returned `error.Interrupted`** - which `request()`'s retry loop deliberately does NOT retry (Esc semantics) - while the non-streaming `postWatched` correctly returns retryable `error.HungRequest`. So even a *detected* stall ended the turn instead of reconnecting.

## Done (release/0.0.16)

- Guarded `peekByte` with the same stall watchdog (new `streamPeekTask` select-arm).
- Watchdog deadline stalls now return `error.HungRequest` (retryable -> `request()` reconnects + redials a fresh connection); real Esc still maps to `error.Interrupted`.
- `zig build` + `zig build test` green.

Net: a dead/half-open stream now trips the watchdog (<= `stream_stall_ms` = 120s), prints `[network error: HungRequest - retrying]`, and redials (up to 3 attempts) instead of hanging at "thinking".

## Remaining - the broader watchdog work

1. **`receiveHead`: guard the HTTP response-header read.** Before the line/peek loop, `postStream` reads the response head. A server that accepts the TCP connection but never sends headers (or sends them then dies) hangs *there*, before the first byte - same "stuck at thinking", not yet covered. Race `receiveHead` (and/or the request send) against the stall/deadline watchdog so a pre-stream hang also becomes a retryable `HungRequest`.
[...]

PR #215, opened 2026-07-14
fix(agent): don't peek past a completed stream on a half-open socket (#56)

justrach:
## Problem

The SSE streaming reader's post-line probe, `reader.peekByte()` (`agent_stream.zig`), ran **before** the terminal-event check that breaks on `[DONE]` / `response.completed`. On the final line of a response the peek's result is discarded anyway (the loop breaks at the terminal check), but its *fill* can block **forever** on a half-open socket — server gone, no FIN/RST — which the idle-stall watchdog does not guard. That hangs an already-complete turn: part of #56's "reconnect doesn't reconnect".

## Fix

Move the peek to **after** the terminal-event break, which matches this loop's own documented intent — *"stop instead of waiting for the socket to close"*. A completed response now breaks before it ever peeks; the non-terminal path is byte-for-byte unchanged (`peek → if (!more) break → toss the buffered '\n'`).

**Safe by inspection:** on a terminal line the loop already broke at the `isStreamEnd` check, so `more` was never read there — moving the peek only removes a useless, hang-prone fill on completed responses.

## Verification

- `zig build test` green.
- Streaming PTY suite green — `test-pty-codex-ws`, `test-pty-repl`, `test-pty-markdown`, `test-pty-spinner` (these stream model output through this exact reader), so **no regression in the hot read path**.

I explored a mock hang-test but it can't reliably distinguish old-from-new: a complete SSE line always leaves its terminating `\n` buffered (that's the byte `reader.toss(1)` consumes), so the old `peekByte` reads from the buffer without blocking in every reproducible case. A test both versions pass isn't a meaningful guard, so I rely on the inspection argument + regression suite instead.

## Scope (honest)

- **This PR** closes the **post-completion** subcase (the documented "don't wait on the socket" intent).
- The **mid-stream** half-open subcase — `peekByte` blocking on a *non-terminal* line — still needs a watchdog-guarded peek (a `streamPeekTask` select-arm) **plus** a half-open-socket mock harness (accept, send partial SSE, then network-drop without closing) to verify. Left as a follow-up.
- **Fix-B** from the issue (retryable stalls via `watchdogError`: Esc→`Interrupted`, deadline→`HungRequest`/`StreamStalled`) is already implemented and tested (#134).

Pre-existing `zig fmt` debt in this file (unrelated one-liners) left untouched to avoid churn.

Refs #56

### DIFF
--- a/src/agent_stream.zig
+++ b/src/agent_stream.zig
@@ -398,9 +398,6 @@ pub fn postStreamWithClient(self: *Agent, client: *std.http.Client, body: []cons
                 },
             }
         }
-        // End of stream leaves the reader empty; otherwise the '\n' is
-        // still buffered (and consumed below, after the line is handled).
-        const more = if (reader.peekByte()) |_| true else |_| false;
         try full.writer.writeAll(line.writer.buffered());
         try full.writer.writeByte('\n');
         got_body = true; // #134: response bytes are in `full`; a later read error is a clean close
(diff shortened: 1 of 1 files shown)


justrach:
Already on main (half-open socket peek guards in agent_stream / agent_ws). Closing as superseded.

PR #497, opened 2026-08-13
feat(sdk): add production embedding parity

justrach:
## What changed

- Adds production-safe TypeScript embedding primitives: serialized local/remote operations, out-of-band `answer()`/`cancel()`, `AbortSignal`, provider-qualified controls, explicit compaction, structured usage/results, and reliable async lifecycle handling.
- Gives `--json` one stdin owner so requests, ask replies, and cancellation cannot steal each other's lines; serve cancellation bypasses the busy gate and leaves sessions reusable.
- Packages Graff as six platform-specific optional npm packages and resolves binaries in this order: explicit path, compatible native package, then the caller's child-process `PATH`.
- Validates the native package's `graffProtocol` before spawn, keeps generation authoritative in `sdk/generate.py`, and adds unit, protocol, clean packed-consumer, and publishing-workflow coverage.

### Problem / failure mode

The SDK exposed basic local and remote chat, but production embedders could race operations, lose events, hang during shutdown, or fail to cancel a turn cleanly. An npm consumer also needed a separately installed `graff` executable, so installing `@codegraff/sdk` alone was not enough to run.

### Reason for this approach

A single JSON inbox establishes clear ownership of stdin while preserving out-of-band replies and cancellation. The SDK serializes stateful operations at its boundary and reports terminal usage as a structured result. Esbuild-style optional native packages make normal npm installation self-contained without executing download code during install, while protocol validation prevents an SDK from silently starting an incompatible binary.

- There are six native packages, increasing release workflow surface area, but installs download only the matching optional dependency.
- Native publication intentionally requires an explicit `GRAFF_NATIVE_RELEASE` whose binary reports protocol `0.12`; the existing `v0.0.251` assets report `0.11` and must not be used.
- Explicit binary configuration remains supported and takes precedence; `PATH` remains a fallback for development and unsupported package-install cases.
- This does not add Pi's session-tree/plugin/TUI ecosystems and does not change the GUI.

### Rejected alternatives

- Postinstall downloads were rejected because they add network execution and platform failure modes during package installation.
- Bundling every platform binary into the main SDK was rejected because it would make every install unnecessarily large.
- Falling back to an obsolete `harness` executable name or accepting a protocol mismatch was rejected because both can launch the wrong program and fail later in less actionable ways.

## Verification

- `bash scripts/eval-tier1.sh` — green; 1,207 Zig tests, reachability and SDK drift checks included
- `bun test` — 34/34
- `npm run typecheck`
- `npm run build`
- `npm run test:packed`
- `npm run test:packed -- --binary ../../zig-out/bin/graff`
- `python3 scripts/test-json-controls.py zig-out/bin/graff` — 21/21
- `python3 scripts/test-json-cancel.py zig-out/bin/graff`
- `python3 scripts/test-json-live.py zig-out/bin/graff`
- `actionlint .github/workflows/ci.yml .github/workflows/sdk.yml`


justrach:
Hosted CI status: both SDK jobs and both Windows jobs pass. Both duplicate Zig jobs reach the existing PTY `/goal` check and then fail because startup cosmetic model calls make it observe 7 requests instead of 4.

I reproduced the same `/goal status ran a model turn` failure on an untouched detached `origin/main` (`cb27ab2`), so it is not introduced by this PR. The existing `fix/repl-startup-diagnostics` branch addresses that REPL startup behavior; I intentionally did not fold those unrelated startup changes into this SDK-focused diff.

Issue #472, opened 2026-08-08
Foreground bash gives zero output until exit, so long commands look like the harness has frozen


justrach:
## What happens

A foreground `bash` tool call returns nothing at all until the process exits. For a command that takes minutes — a test suite, a build, a deploy — the session goes completely silent, with no output, no elapsed time, and no indication anything is alive. From the outside it is indistinguishable from a hang, and it blocks the whole turn: nothing else can happen while it runs.

Reported by a user mid-session as "why does codegraff always hang here tho" and "it js stops everything".

## Evidence from one real session

Harness `v0.0.243-4-g18e3f31-dirty`, measured from one session's local event trace:

- **9** `bash` calls over 60s
- **4302s (72 min)** of wall time inside those 9 calls
- **5120s (85 min)** total tool wall time across 198 calls

So ~84% of all tool time was spent in blocking calls that displayed nothing while running. Individual offenders:

| ms | what it was |
|---|---|
| 2977129 (50 min) | a vendor CLI's log-streaming subcommand — streams, never exits, had to be cancelled by the user |
| 355904 (5.9 min) | `vitest run` against a remote database |
| 260087 | same suite |
| 259741 | same suite |
| 163511 | `vitest run` single file |

## Why it is worse than "just slow"

1. **No partial output.** The command is producing output the whole time; none of it is shown. The buffer is only flushed on exit.
2. **Streaming commands never exit at all.** A log-streaming CLI subcommand, `tail -f`, a dev server — these have no exit. In the foreground they consume the turn until a human notices and cancels. The 50-minute call above is exactly this.
3. **The documented workaround is invisible at the call site.** `run_in_background` + `bash_output` exists and works, but nothing about a foreground call signals "this one is going to be long" until it is already too late. An agent only learns after burning the time.
4. **Piping defeats the workaround too.** Backgrounding a job but piping through `tail -40` reproduces the silence, because `tail` buffers to EOF — so `bash_output` returns "(no new output)" for minutes on a job that is working fine. Easy trap to fall into.

## Suggested fixes, roughly in order of value

- **Stream partial stdout/stderr for foreground bash** as it arrives, even truncated to the last N lines, instead of buffering to exit.
- **Auto-promote to background after a threshold** (say 30-60s): return a job id and the output so far, with a note that it kept running, rather than holding the turn open indefinitely.
- **Emit a heartbeat** — elapsed seconds and bytes-so-far — so a silent-but-working command is distinguishable from a wedged one.
[...]

Issue #462, opened 2026-08-06
Windows: write-only file handles + File.stat silently break appends (playbook inert, serve event log overwrites from 0)


justrach:
Found while fixing #441's Windows CI failure (PR #457). The same defect exists in at least two other places, both unfixed, both on a platform graff has supported since v0.0.167.

## The mechanism

`Io.Dir.createFile` maps `read` straight onto the NT access mask (`.GENERIC = .{ .WRITE = true, .READ = flags.read }`). A write-only handle therefore carries `FILE_GENERIC_WRITE`, which does **not** include `FILE_READ_ATTRIBUTES`. `File.stat` is `NtQueryInformationFile(.All)`, which requires exactly that right, so **`stat` on a write-only handle fails with `ACCESS_DENIED` on Windows** and succeeds on POSIX.

Any code that opens a file write-only and then stats that handle to find the append offset is silently broken on Windows. It is invisible on macOS and Linux, including in the full test suite: #441 reverted its fix and still ran 1050/1050 green locally. CI's Windows job is the only gate for this class.

## Instance 1: `playbook.zig:appendLine` — playbook appends are lost

Identical write-only-handle plus `f.stat` pattern. On Windows the stat fails, the append is abandoned **after the file has already been created**, and the playbook silently records nothing. Given #383's playbook is the substrate for behavioural curation and rides every subagent brief, a Windows user gets a permanently empty playbook with no error.

## Instance 2: `serve_events.zig:58` — every event overwrites from offset 0

This one already *works around* the failure rather than fixing it:

```zig
if (file.stat(io)) |st| st.size else |_| 0
```

On Windows that `else` branch is taken **every time**, so the computed append offset is always 0 and each event overwrites the log from the start. The durable EventLog is therefore corrupt on Windows: no gap-free `?from=N` replay, no usable Follower tail.

**This matters for #431.** Detachment milestone 1 is explicitly built on serve's "seq-stamped EventLog, gap-free `?from=N` replay, and Follower tailing", and its acceptance criterion is an attach-parity eval plus "kill the supervisor mid-session: the worker survives and a reattach replays without gaps". On Windows that replay cannot work today. Worth fixing before #431 rather than discovering it in the attach eval.

## Fix

`.read = true` on the create, plus a path-based `statFile` fallback so a future std change degrades to a working append rather than a silent no-write. That is what #441 landed in `session_transcript.zig`; the same shape applies to both sites above.

## Test guidance

[...]

Issue #452, opened 2026-08-06
No GC sweeper for .graff/tool-results/ — handle artifacts accumulate for the life of a workspace


justrach:
Surfaced by the #440 implementation (PR #451), and pre-existing rather than caused by it.

**Today**: `.graff/tool-results/` accumulates one file per oversized tool result and nothing ever removes them. Unlike `tool_spill`'s session artifacts, which #409 gave a sweep tied to the session lifecycle, this directory has no cleanup path at all: not on session end, not on `/new`, not on exit.

**Why it matters more now.** #440 makes handles the default contract rather than an overflow rescue, so the write rate goes up, and it raised the bash capture ceiling from 128 KiB to 1 MiB, so individual artifacts get larger. #440 added a 64 MiB **per-process** budget with an honest truncation marker, which bounds any single run, but that budget resets every process. A long-lived workspace accrues without limit across runs.

**Proposal**: give the directory the same lifecycle treatment `.graff/` artifacts already got in #409, rather than inventing a second cleanup mechanism.

- Sweep on the same trigger as the #409 session sweep.
- A workspace-level cap (total bytes or age), enforced by evicting oldest-first, since a handle's value decays sharply once its session is gone.
- Decide what a resumed session is owed: #440 deliberately keeps the send-time backstop alive for sessions resumed from a pre-#440 build, so an over-eager sweep could delete artifacts a resumed session still points at. Eviction should not orphan a live reference.

**Interaction**: #441 (append-only transcript) and #411 (post-compaction note listing live handles) both make handles longer-lived by referring to them after compaction. Whatever retention rule this lands on should be the one those two cite, not a third policy.

justrach:
Shipped in #511: `.graff/tool-results/` orphans are swept with the same session-lifecycle pass as the #409 spill sweeper.

Issue #409, opened 2026-08-06
Spill, don't truncate: oversized tool outputs go to a session artifact, cap note cites the path


justrach:
From the prime-agent analysis (PrimeIntellect-ai/prime-agent, 2026-08-06). Their model keeps every large result addressable (a Python object in the kernel namespace) and only caps what reaches the transcript (64KB/cell). The transferable idea needs no kernel: never destroy data at the cap point.

**Today**: `capOversizedToolOutputs` (#196) caps an oversized tool output per-output when over the context budget. The elided content is gone; the model re-runs the tool if it guessed the wrong slice.

**Proposal**: before capping, write the full output to a session artifact (e.g. `.graff/sessions/<id>/artifacts/tool-<call>.txt`) and make the elision marker actionable: `[truncated: full output (N bytes) at <path>]`. The model can then read/grep exactly the slice it needs.

**Notes**:
- Same call sites as #196 (anthropic/openai/codex paths); same safety class as the cap itself (no WS-close bracketing needed, per the #196 analysis).
- Bound artifact growth: per-session byte cap, artifacts deleted with the session.
- Only spill when a durable session exists; `--no-session`-style runs keep plain truncation.

justrach:
Shipped in v0.0.241 (merge train PR #446; released 2026-08-06). Verified by golden evals, CI, and live gpt-5.6-sol runs — see the release notes and the live-eval findings referenced in #440/#445 for measured impact.

PR #511, opened 2026-08-15
fix: stream foreground bash and close the small follow-up batch

justrach:
Foreground bash used to stay silent until the child exited, so a long `zig build test` looked like a hang. This streams new stdout/stderr every 200ms into the live pane (and as `tool_output` JSON).

Also the rest of the small/medium list, nothing from the epic pile:

| Issue | What landed |
| --- | --- |
| #472 | `runCapped` stream callback + TTY/TUI live pane + JSON `tool_output` |
| #406 | failed `/rewind` write-backs increment `skipped`; pin-test scans sibling emits |
| #452 | `.graff/tool-results/` GC on the same session-lifecycle sweep as #409 |
| #404 | refuse clobbering unreadable `auth.json`; skip adopt when persist failed |
| #405 | `replaceFile` writes through a symlink instead of replacing it |
| #320 | `.graff/owner.json` + startup warning when another live root owns the worktree |
| #388 | playbook reflector sits on the root provider, not a worker pin |
| #249 leftover | first-class `view_image` tool (advertised when the session model is a VLM) |
| #108 leftover | `@[path]` composer lines stay in input history |
| #462 | playbook append + serve event log open write+read and fall back to `statFile` |

`zig build test` 1292/1292. Tier 1 green. SDKs regenerated for `view_image`.


justrach:
Closing as landed. Foreground bash streams via `exec_bash_stream` / `tool_output_delta` (and auto-backgrounds at 120s, ADR 0026). The rest of this follow-up batch (`/rewind` skipped, tool-results GC, `auth.json`, symlink `replaceFile`, worktree owner, `view_image`) is on tip.

Issue #602, opened 2026-08-21
Harness observability: /trajectory cannot show what the model did


justrach:
## Summary

`/trajectory` is a DGM fitness ledger (prompt SHA, recipe SHA, parent, score). That job is correct and should stay. It is **not** a replay of what the model did, and the other local streams do not fill that gap either.

Compared with grok-build's session folder (`events.jsonl` + `chat_history.jsonl` + live `updates.jsonl`) on the same model (grok-4.6, effort high), a real graff session is unreadable as a trajectory. The data is split across three files; two of them lie or lag; `/trajectory` prints almost none of what they do contain.

Related but distinct: ADR 0019 (transcript shows decisions) and #246 (behavioral schema, closed). This issue is the **inspect path** — can a human, after or during a run, see the movie.

## Evidence (run `28d9d52384b3fc6a`, 2026-08-21, grok-4.6 / xAI / 0.0.267)

Session: merge `release/v0.0.268` into `main`, pull/codesign `0.269`.

| Stream | What it wrote |
| --- | --- |
| `.graff/trajectories/28d9d52384b3fc6a.jsonl` | 3 closed turns. Turn 4 (34 API / 98 tools, the live work) never landed. |
| `.graff/traces/28d9d52384b3fc6a.jsonl` | ~500 `api`/`tool`/`ws` events with `name/ms/result_bytes/is_error`. No args, no call id, no first_token. |
| `.graff/behavior/28d9d52384b3fc6a.jsonl` | 1 KB. `run_started` + `turn_started` 1–5. No `turn_committed`, no `run_finished`, no tool events. |
| `.graff/sessions/merging-0-268-then-pulling-0-269.transcript.jsonl` | Full Responses items (args + outputs + assistant text) — closest to grok-build `chat_history.jsonl`. `/trajectory` never reads it. Last flushed at turn 3. |

Turn 1 DGM row:

```
tools: "bash,bash,bash,bash,bash,bash,bash!,bash,bash,bash"
ok: true, ms: 95146, model_calls: 10
resp_blocks: 0, resp_text_len: 0, resp_types: ""
cost_microusd: 0
```

The same turn's `recipe_outcome` recorded `tool_calls: 14`. The extra four are meta (`todo_write`, `peer_message`, `load_tool_schemas`, `attempt_completion`) and never enter `ToolSink`. The transcript has the assistant narration; the turn node says the model produced nothing.

[...]

Issue #601, opened 2026-08-21
Line REPL tool rows: remaining rendering after ADR 0019


justrach:
## Problem

ADR 0019 made tool rows one interpreted line instead of live bash / JSON envelopes. That part works:

```text
  ✓ read  src/version.zig  full · 72 lines
  ✓ read  docs/architecture.md  lines · 161 lines
```

What is still wrong:

1. **`mode=lines` is tautological.** `lines · 161 lines` prints the CodeDB mode name as if it were a decision. `full · 72 lines` is only slightly better. The user-relevant fact is `architecture.md · 161 lines`.
2. **The wall of prose still follows.** After those two rows the model dumps a full architecture essay (layout, branch vs main, how the watcher works). The chrome cannot stop model narration; there is no fold/expand on the line REPL, so the transcript still is not something you can stare at all day.
3. **No progressive disclosure.** Raw output is `/debug` or the TUI fold only. Enter/click to expand a one-liner was part of the original sketch and is not built.
4. **JSON interpretation is still mode-leaky.** `interpretJson` surfaces `mode` verbatim (`section`, `outline`, `lines`, `full`) instead of a noun the user would miss (`161 lines`, `outline`).

## Not this issue

Do not bring back live bash dumps, command-as-detail, or the `{"ok":true,"file":...}` envelope. Those are the regressions ADR 0019 exists to prevent.

## Wanted

```text
  ✓ read  version.zig · 72 lines
  ✓ read  architecture.md · 161 lines
```

and a way to open the raw result on demand. Model essays are a separate prompt-policy question; this issue is the harness chrome.

justrach:
The mode-leak half is fixed in #642: `interpretJson` now interprets a content-bearing read as `161 lines` — never `lines · 161 lines` or `full · 72 lines`; `outline`/`symbol` stay nouns. Row reads `✓ read architecture.md · 3 lines`.

Still open from this issue: progressive disclosure (enter/click to expand a one-liner). Raw output remains at `/debug` and spilled results badge `⇢ N KiB handle`, per ADR 0019.

justrach:
[...]

Issue #270, opened 2026-07-22
Agent returns unrelated Chinese refusal during benign coding task


yxlyx:
## Bug

During a long portfolio redesign session, the agent abruptly returned an unrelated Chinese refusal:

> 你好，我无法给到相关内容。

The preceding task was benign: finish serving two locally created HTML portfolio mockups and present their URLs. There was no safety-sensitive request and no reason to switch language or refuse.

## Expected behavior
The agent should finish the benign coding/design task and present the preview URLs, or surface a concrete API/tool error.

## Actual behavior
The response suddenly switched to Chinese and said it could not provide the content. The user had to type `continue` before work resumed normally.

## Trace evidence and limitation
The API call near the event completed successfully (`is_error: false`, about 57k context tokens), but `harness.trace.jsonl` logs API/tool metadata—not final assistant text. Therefore the trace alone cannot distinguish an empty response from a text-only response. The original issue incorrectly characterized this as a silent empty turn; the user-visible transcript confirms it was an unrelated Chinese refusal instead.

## Context
- Long web-development redesign session
- Approximately 57k context tokens near the anomalous response
- Two parallel mockup-building subagents had just completed
- Benign local HTML/CSS work
- Retrying via a user `continue` message worked normally

## Suggestions
- Preserve/log final-response content-block metadata (at least block count/type/language, without necessarily logging sensitive text) so empty and text-only responses can be distinguished.
- Detect abrupt unrelated refusals/language switches on benign tool-driven tasks and retry once or surface a diagnostic.
- Investigate whether long-session compaction or response assembly can substitute/truncate the intended final response.

justrach:
Partially addressed in v0.0.226 (PR #328) - **leaving this open**.

Landed: The turn trace now records the final response's block SHAPE (block count, which block types were present, total text length), so a text-only refusal is distinguishable from a turn that did nothing. No content is recorded.

Still open: refusal/language detection and the retry-or-steer behaviour were deliberately not attempted - a heuristic there would be worse than nothing.

justrach:
Closing: the actionable half shipped in #328 / v0.0.226 — `src/mainloop_trace.zig` records final-response block shape (tests tagged #270), so a text-only refusal is now distinguishable from an empty turn — and the remaining ask (refusal / language-switch heuristics with auto-retry) was deliberately declined as worse than nothing. No repro in over a month; reopen with a trace if it recurs.

PR #642, opened 2026-08-26
fix(#601,#602): row nouns without mode names; trajectory shapes and shows what turns did

justrach:
## What changed

- **#601** — `interpretJson` no longer surfaces the CodeDB mode name verbatim. A content-bearing read now interprets as `161 lines` instead of `lines · 161 lines` (tautology) or `full · 72 lines`; modes that are themselves the fact (`outline`, `symbol`) stay.
- **#602 (2 of 5 slices)** —
  - `finalResponseShape` broke its backward walk on a trailing bare `function_call_output`, so every tool-using turn whose history ends on tool results (turn cut after execution, crash, attempt_completion's final call) archived `resp_blocks=0` — indistinguishable from an empty turn. Those items are now skipped; the shape describes the last model-authored run.
  - `/trajectory` printed only `turn N ✓ 95s · prompt <sha> · <task>`. The row now appends what the node already carried: model calls, tool count, tool errors, cache-hit percent, and response shape (`text+reasoning`), when present.
  - `commands_session.zig` crossed the 600-line ceiling with the above; the `/trajectory` handler moved verbatim into `commands_trajectory.zig`.

## Why

The recorder bug made the DGM archive lie about tool-using turns: #270 added `resp_*` precisely to separate 'text-only refusal' from 'empty turn', but the Responses+tools path defeated it, so a history ending on tool results was never covered by a test. And `/trajectory` is the human inspect path — it ignored fields already on the turn node, leaving a run unreadable compared with a grok-build session folder.

## Out of scope (kept open on #602)

Live turn stubs in the DGM file, periodic session-transcript flush, and `first_token`/`call_id` on operational traces. Behavior lifecycle (`turn_committed`/`run_finished`) already landed earlier. Click-to-expand for line-REPL rows (#601's progressive disclosure) stays out per ADR 0019 — `/debug` remains the raw path.

## Verification

- New tests: mode-noun cases in `agent_tool_label.zig`; cut-after-execution shaping case in `mainloop_trace.zig`. Updated the one render test that asserted the old `section · N lines` string.
- `zig build test` green; `scripts/eval-tier1.sh` green locally (twice) and via the pre-push hook.

### DIFF
--- a/src/agent_tool_label.zig
+++ b/src/agent_tool_label.zig
@@ -357,9 +357,14 @@ fn interpretJson(s: []const u8, buf: []u8) []const u8 {
     if (jsonQuoted(s, "matches")) |m|
         return std.fmt.bufPrint(buf, "{s} matches", .{m}) catch m;
     if (jsonQuoted(s, "mode")) |mode| {
+        // #601: the CodeDB mode name is not the decision. A read that carried
+        // content says how much came back (`161 lines`) — the row's detail
+        // column already names the file, so `lines · 161 lines` was a
+        // tautology. A mode that IS the fact (`outline`, `symbol`) stays.
         const lines = if (jsonQuoted(s, "content")) |c| escapedNewlineCount(c) else 0;
         if (lines > 1)
-            return std.fmt.bufPrint(buf, "{s} · {d} lines", .{ mode, lines }) catch mode;
+            return std.fmt.bufPrint(buf, "{d} lines", .{lines}) catch mode;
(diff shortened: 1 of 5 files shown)


justrach:
Closing as landed on tip. Row nouns without mode names (`161 lines` not `lines · 161 lines`) and `finalResponseShape` skipping a trailing bare `function_call_output` are already in `agent_tool_label.zig` / `mainloop_trace.zig`. Dirty vs main; no unique leftover.

PR #664, opened 2026-08-28
fix: TUI first paint no longer waits on companion MCP

justrach:
## What you saw

```
graff tui --yolo
plugins: 16 plugin(s) in 1ms (7 dirs)
```

Then nothing for a long time. Plugin scan was already fine. The hang was the **next** work, with no receipt.

## Where the time went

After `plugins:`, interactive `--yolo` still did two things on the main thread before the alt screen:

1. **Companion auto-connect.** Deferred MCP leaves `registry.tools` empty, so a PATH-installed `codedb-pro` looked “not connected” and `addServer` ran the full stdio probe + handshake (5s / 15s caps) in front of first paint.
2. **`codedb-pro probe` with timeout 0.** If that binary blocks, boot never continues.

A `defer_join` fan-out that could not `io.concurrent` also fell back to `io.async`, which can run the handshake inline.

## What we did

- Print dim receipts after `plugins:` so the next phase is named:
  - `mcp: N server(s) in background` / `mcp: connecting N server(s)...`
  - `mcp: ready in Xms` if that wait is ≥80ms
  - `companion: codedb-pro (background)`
- Queue companion onto `pending_starts` on interactive `--yolo` (same as ADR 0035 MCP defer). Native tools run now; catalogs merge on the next request.
- Cap `codedb-pro probe` at 2s.
- Skip a server rather than inline-`async` when `defer_join` cannot get a thread.
- Print any boot phase ≥80ms without `GRAFF_BOOT_DEBUG`.

## How it looks

Line REPL (`graff --yolo`) keeps the same dim `plugins:` line and adds two siblings:

```
plugins: 16 plugin(s) in 1ms (7 dirs)
mcp: 3 server(s) in background
companion: codedb-pro (background)
```

`graff tui --yolo` prints those, then takes the alt screen. First paint is the welcome composer — no `plugins:` / `companion:` on the grid.

## Check

`graff tui --yolo` should paint right after the plugin line. You should see `companion: codedb-pro (background)` if the binary is on PATH. MCP tools still arrive on the next turn / `/mcp`.

Opt out of the companion the same way as before: `{"skills": {"codedbpro": false}}`.

Tier 1 green (1762 tests).

### DIFF
--- a/TUI/render_tests.zig
+++ b/TUI/render_tests.zig
@@ -36,6 +36,15 @@ test "welcome frame has chrome, prompt, and no offline stub" {
     try std.testing.expect(std.mem.indexOf(u8, frame, "offline") == null);
     try std.testing.expect(std.mem.indexOf(u8, frame, "Fullscreen") == null);
     try std.testing.expect(std.mem.indexOf(u8, frame, "click") == null);
+    // Boot receipts (plugins:/mcp:/companion:) print before alt-screen.
+    try std.testing.expect(std.mem.indexOf(u8, frame, "plugins:") == null);
+    try std.testing.expect(std.mem.indexOf(u8, frame, "companion:") == null);
+    if (std.c.getenv("GRAFF_DUMP_WELCOME")) |path_z| {
+        const vis = try dump_mod.visible(std.testing.allocator, frame);
+        defer std.testing.allocator.free(vis);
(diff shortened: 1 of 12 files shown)


justrach:
Landed on `release/v0.0.281` (#670) as merge `54601b1`. Closing this vs main so it does not double-land; it ships when 281 merges.

PR #674, opened 2026-08-29
fix(readline): make pasted text placeholders atomic

### LINKED ISSUE #673: TUI: render pasted-text placeholders as highlighted atomic spans
## What changed / requested behavior

Render long or multiline paste placeholders such as `[Pasted text #1 +12 lines]` as highlighted semantic spans in the TUI composer, similar to image/file markers, and treat each placeholder as one logical editing unit.

At minimum:

- A placeholder backed by stored pasted text is visually highlighted, making it clear that the full payload was captured and remains attached.
- Option/Alt+Backspace with the cursor after the placeholder removes the entire placeholder in one action rather than deleting `lines]`, `+12`, `#1`, etc. word by word.
- Word-left/word-right navigation treats the placeholder as one unit rather than entering its display text.
- Removing the semantic span detaches its hidden paste body.
- Manually typing an identical-looking string must remain ordinary text and must not reattach a previously removed hidden payload.
- Multiple pasted spans retain independent identity and expand to their corresponding bodies exactly once on submission.

Scope: terminal TUI/readline behavior only; no GUI changes.

## Why

Long/multiline paste content is currently stored out of band while a plain editable string is inserted into the composer (`src/readline.zig`). Submission finds that exact string and replaces it with the hidden body. This creates two confusing behaviors:

1. The placeholder looks like ordinary text, so users cannot tell whether it still represents intact pasted content or has been partially corrupted by editing.
2. Generic word deletion treats one paste attachment as several words even though it is conceptually one object.

There is also an identity problem: deleting the full placeholder does not remove its backing record during the same prompt. Manually reconstructing the exact marker currently makes submission expand the old hidden paste again. Highlighting alone would not solve that; the composer needs to track a live semantic span/attachment rather than infer attachment identity from matching display text.

Images and selected file paths already establish the visual language for highlighted composer spans. Reusing that language makes a highlighted paste marker mean “this hidden payload is present and intact,” while an unhighlighted lookalike is visibly just text.

## Constraints and trade-offs

- Do not put the full multiline body back into the visible composer; the compact placeholder is still useful.
- Preserve the exact pasted bytes used for submission, including multiline content.
- Styling must follow the existing TUI marker/highlight treatment and remain readable in both terminal themes.
- Avoid relying only on string matching: presentation text is not a stable attachment identity.
- Keep ordinary text editing and short single-line paste behavior unchanged.

## Acceptance checks

1. Paste 12 lines: one highlighted `[Pasted text #1 +12 lines]` span appears.
2. Press Option/Alt+Backspace immediately after it: the complete span and its payload are removed in one action.
3. Delete the span, manually type `[Pasted text #1 +12 lines]`, and submit: the literal string is sent; the removed body is not resurrected.
4. Paste two bodies: deleting one leaves the other highlighted and expands only the surviving payload.
5. Move by words across a paste span: the cursor lands before or after it, never inside it.

yxlyx:
Closes #673.

## What changed

- Added a semantic paste store that owns each collapsed paste body and tracks the byte range of its visible `[Pasted text #N +L lines]` label.
- Rendered only live, position-validated paste spans with the existing reverse-video/emerald attachment treatment.
- Made ordinary and word navigation skip paste spans as one unit; Backspace, forward delete, Ctrl/Option word deletion, line deletion, completion, insertions, and clears now update or detach span identity.
- Expanded only tracked spans on submit and consumed their identity after expansion, so a typed lookalike remains literal and repeated expansion cannot duplicate a body.
- Detached paste identity when history replaces the draft, because history persists display text but not hidden paste bodies.
- Split history replay into `readline_replay.zig` so `readline.zig` remains below the 600-line source ceiling.

### Problem / failure mode

The composer stored long paste bodies out of band but represented them with ordinary editable text. Submission searched for the exact display string and replaced every match. Users therefore had no visual proof that a marker still had an intact payload, Option/Alt+Backspace deleted the marker word by word, and deleting then manually recreating an identical label could resurrect hidden content.

### Reason for this approach

Attachment identity must not be inferred from presentation text. Positional semantic spans let the editor distinguish a live paste from an identical typed string, render that distinction, shift the attachment with surrounding edits, and detach it whenever an edit touches the span. The same ranges also define atomic cursor and deletion boundaries.

- Short single-line paste behavior remains ordinary inline text.
- Full multiline bodies remain out of the visible composer and preserve their exact submitted bytes.
- History replay deliberately drops semantic identity: the history layer has no hidden paste body to restore, so highlighting or expansion there would be misleading.
- This change is limited to the raw terminal readline/TUI path; GUI behavior is unchanged.

### Rejected alternatives

Adding paste labels to the existing string-based `marks` list would highlight every identical lookalike and preserve the resurrection bug. Special-casing word-boundary deletion without semantic identity would improve Option+Backspace but would not establish whether a payload was still attached or make submission safe.

## Verification

- `scripts/eval-tier1.sh`
  - formatting and 600-line ceiling green
  - spec and test reachability green
  - build green
  - 1,760 unit tests passed, 1 skipped
  - 459/459 fullscreen-TUI tests passed
  - all 17 PTY probes passed
  - invariants and SDK synchronization green
- `zig test src/readline_paste.zig` — 8/8 focused state-machine tests passed.
- `zig build test -Dtest-filter='paste' --summary all` — 60/60 relevant tests passed.
- Independent real-PTY verification confirmed:
  - paste and file attachment spans use the same ANSI chip styling,
  - `ESC DEL` (Option/Alt+Backspace) removes the complete paste marker,
  - manually retyping the removed label submits it literally without the hidden body.
- Independent edge-case passes covered multiple spans, deletion/offset shifts, lookalike labels, history replay, monotonic IDs, and an adversarial body beginning with its own display label expanded twice without duplication.

### DIFF
--- a/src/input_util.zig
+++ b/src/input_util.zig
@@ -23,6 +23,7 @@ const runCapped = jobs.runCapped;
 
 const pricing = @import("pricing.zig");
 const util = @import("util.zig");
+const readline_paste = @import("readline_paste.zig");
 
 const main_mod = @import("main.zig");
 const provider_mod = @import("provider.zig");
@@ -303,12 +304,12 @@ pub fn cleanDroppedPath(gpa: Allocator, home: []const u8, pasted: []const u8) ?[
 
 // Redraw the whole input below a fixed prompt prefix, wrapping it
 // across rows. Spans listed in `marks` (paths from the @ picker or a
-// file drop, plus "[Image]") render as an accent chip (reverse video +
-// Codegraff emerald) so they keep reading as attached files, not typed words; a
-// chip crossing a row break keeps its colour. `st` carries the row
-// count + cursor row of the previous draw so this one can clear it
-// with relative moves only — no DECSC anchor for a scroll to strand.
-pub fn redraw(o: *Io.Writer, items: []const u8, c: usize, marks: []const []const u8, st: *LineRender, pcol: usize) void {
+// file drop, plus "[Image]") and live semantic paste spans render as accent
+// chips (reverse video + Codegraff emerald), so attachments do not look like
+// typed words. A chip crossing a row break keeps its colour. `st` carries the
+// row count + cursor row of the previous draw so this one can clear it with
+// relative moves only — no DECSC anchor for a scroll to strand.
+pub fn redraw(o: *Io.Writer, items: []const u8, c: usize, marks: []const []const u8, pastes: ?*const readline_paste.Store, st: *LineRender, pcol: usize) void {
     const cols = termCols();
     const plen = if (pcol > 0) pcol - 1 else 0; // columns the prompt holds on row 0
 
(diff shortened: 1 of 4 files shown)


justrach (approved):
Checked against #673 acceptance on the current `main` composer (string-match expand) vs this branch.

This is the right fix: identity lives on a positional paste store, highlight is only for live validated spans, Option/Alt+Backspace and word-nav treat the marker as one unit, submit expands only tracked spans, and a typed lookalike stays literal. History dropping semantic identity is also correct — there is no hidden body to restore.

CI is green (zig/sdk/windows). No follow-up needed from this review; leaving #673 to close with the merge.

justrach:
Folded into release/v0.0.282 (#696). Closing as superseded.

PR #693, opened 2026-08-30
fix(codex): preserve terminal WebSocket API errors

### LINKED ISSUE #692: codex: generic Bad Request frame is misclassified as WS transport failure; next continue succeeds
## Problem / failure mode

Codex tool-heavy turns repeatedly end with `codex api error: Bad Request`. Re-prompting with `continue` succeeds immediately, but graff spends a WS redial and SSE fallback first and loses the useful error-frame detail.

This is a TUI runtime/provider-path issue, not presentation-only behavior.

## Evidence

Six independent trajectories contain this same failure, always on the first root turn; recorded follow-up `continue` turns succeed:

- `.graff/trajectories/8997e379e8f1e9a6.jsonl:6-9`
- `.graff/trajectories/a3334f40111960cd.jsonl:6-9`
- `.graff/trajectories/78e55567de819be8.jsonl:6-9`
- `.graff/trajectories/21cc986aa0d5f15a.jsonl`
- `.graff/trajectories/5e7420d48dfb2b6d.jsonl`
- `.graff/trajectories/188d2d8042e4e325.jsonl:6-9`

The wire sequence is effectively identical across request sizes from ~38 KB to ~149 KB. The current run shows it directly at `.graff/traces/188d2d8042e4e325.jsonl:84-104`:

1. A prior request completes successfully.
2. The next request reuses the WS and receives a first frame.
3. Graff labels the close as a transport error and redials a fresh WS with the same full body.
4. The fresh WS also receives a first frame and closes.
5. Graff permanently falls back to SSE for the session.
6. SSE rejects the identical request with a 25-byte response and the turn ends as `codex api error: Bad Request`.
7. The user's next prompt succeeds on the same process/session/history.

The same sequence appears at `.graff/traces/8997e379e8f1e9a6.jsonl:122-132`, `.graff/traces/a3334f40111960cd.jsonl:42-52`, and `.graff/traces/78e55567de819be8.jsonl:49-59`.

## Why this needs a graff fix

`postResponsesWs` appends every frame to the accumulated response, but `errorFrameAction` only recognizes two specific error codes (`websocket_connection_limit_reached` and `previous_response_not_found`). A generic `{"type":"error",...}` frame is not terminal, so graff waits for another frame; when the server closes, the accumulated API error is discarded and the event is treated as a transport reset (`src/agent_ws.zig:579-599`, `src/agent_ws_signal.zig:135-142`).

That explains the misleading WS retry/fallback and lost diagnostics. The fact that a fresh full-history WS and SSE both reject the exact same body shows this is a request/API rejection, not a flaky socket. The remaining root cause—invalid request shape versus a transient Codex backend rejection—cannot be determined because the generic WS error frame and 25-byte SSE body are not preserved in the trace.

## Expected behavior

- Treat any terminal Codex WS `type:error` frame as an API response and preserve a bounded, redacted code/message in the trace/`last_api_error`; do not call it a transport failure merely because the server closes after the error frame.
- Once the actual code/message is visible, identify and fix the malformed or stale request shape. If adding a bare user `continue` is sufficient to recover, graff should perform the equivalent safe re-anchor/rebuild automatically rather than fail the turn.
- Do not retry deterministic 400 bodies over two WS connections plus SSE without changing the request.

## Regression coverage

Extend the Codex WS mock to send a generic error frame and close. Assert that graff returns the API error once, retains bounded diagnostics, does not burn the transport-failure ladder, and that any recognized recoverable chain error rebuilds full input automatically. Exercise this through the fullscreen TUI simulator/PTY path as well as the lower-level WS parser.

yxlyx:
## What changed

- Treat every authoritative Responses WebSocket `type:error` frame as a terminal API response, retain its accumulated SSE-shaped body, and mark the server-closing socket dead instead of waiting for EOF.
- Parse and expose a single bounded, single-line provider/code/message diagnostic in `last_api_error` and the default trace without copying the raw frame envelope, request, headers, or auth-adjacent fields.
- Rebuild full input exactly once when an error rejects an existing `previous_response_id` chain; deterministic generic API errors return immediately without consuming the WS retry/SSE fallback ladder.
- Preserve that diagnostic through the fullscreen TUI turn bridge rather than replacing it with the misleading generic API-key failure row.
- Extend the loopback WS mock and interactive PTY suite with immediate-close generic-error and stale-chain trajectories, plus headless fullscreen TUI recovery coverage.

### Problem / failure mode

Codex sends a useful `type:error` frame and then closes the WebSocket. Graff accumulated the frame but recognized only two special error codes as terminal. For a generic error it waited for another frame, interpreted the expected close as transport loss, discarded the API body, redialed a fresh WS, and finally replayed the unchanged deterministic request over SSE. Users saw only `Bad Request`, the trace lost the actual code/message, and the session unnecessarily latched off WS. The fullscreen TUI then replaced API failures with an API-key troubleshooting row.

### Reason for this approach

The error frame is an authoritative provider response, so it belongs on the existing Responses parser/API-error path rather than the transport ladder. That keeps one shared parser for WS and SSE, preserves structured recovery such as stale-chain detection, and gives diagnostics one explicit privacy/size boundary before they reach traces or UI. Capturing whether a chain existed before retiring the dead socket allows a safe one-shot full-input rebuild without retrying generic 400 bodies.

- Error frames are parsed only after a cheap `error` candidate check; this adds allocation only on rare candidate frames and avoids brittle substring classification when JSON contains whitespace or quoted prose.
- The diagnostic intentionally keeps only bounded provider, code, and message fields. Additional raw envelope detail is omitted to avoid leaking echoed request/auth material.
- A stale-chain error on a fresh full-input request is surfaced instead of retried indefinitely. Automatic recovery is limited to a request that actually used an existing chain.
- The fullscreen UI regression uses the repository's headless `TUI/sim.zig` path; the provider/socket half is exercised independently through the real interactive PTY runtime and loopback backend.

### Rejected alternatives

- Retrying every generic 400 over another WS or SSE was rejected because the request body is unchanged and the issue traces show that ladder deterministically fails.
- Waiting for a WebSocket close frame was rejected because Codex closes the transport after the error event; EOF is expected protocol aftermath, not evidence that the API response was lost.
- Logging the raw error frame was rejected because envelopes may contain request or authentication-adjacent data and can be unbounded.
- Injecting a synthetic user `continue` was rejected because it changes conversation semantics; a recognized stale chain can be recovered safely by rebuilding the same full input instead.

## Verification

- `scripts/eval-tier1.sh` — green: format, line ceiling, spec, reachability, build, 1,760 tests passed (1 skipped), 460/460 TUI tests, 17/17 TUI guard probes, invariants, SDK.
- `python3 scripts/test-pty-codex-ws.py ./zig-out/bin/graff` — all eight healthy/retry/fallback/error/chain/compaction trajectories passed.
- Five independent subagents separately validated frame parsing edges, immediate-close classification, stale-chain recovery and request counts, fullscreen simulator recovery, and transport/compaction regressions.

Fixes #692

### DIFF
--- a/TUI/sim.zig
+++ b/TUI/sim.zig
@@ -20,10 +20,12 @@
 const std = @import("std");
 const app = @import("app.zig");
 const dump = @import("dump.zig");
+const engine_mod = @import("engine.zig");
 const key_mod = @import("key.zig");
 const keys = @import("keys.zig");
 const render_mod = @import("render.zig");
 const theme_mod = @import("theme.zig");
+const turn = @import("turn.zig");
 
 const Model = app.Model;
 const Effect = app.Effect;
@@ -219,6 +221,41 @@ test "typeText lands in the composer dump" {
     defer std.testing.allocator.free(vis);
     try std.testing.expect(std.mem.indexOf(u8, vis, "hello") != null);
     try std.testing.expect(std.mem.indexOf(u8, vis, "›") != null);
+}
+
+test "#692: terminal API diagnostics survive job finish and the composer recovers" {
+    const a = std.testing.allocator;
+    var term: Term = undefined;
+    term.init(a, 80, 24);
+    defer term.deinit();
+
+    try term.model.push(.user, "trigger a provider rejection");
+    try term.model.push(.pending, "");
+    const job = try a.create(engine_mod.Job);
+    job.* = .{
+        .gpa = a,
+        .history = &.{},
+        .params = .{},
+        .stream = .{},
+        .raw = .{},
+        .threaded = false,
+        .result = try a.dupe(u8, "codex api error [invalid_request_error]: mock bad request"),
+    };
+    job.events.attach(a);
+    job.done.store(true, .release);
+    term.model.pending = job;
+    turn.finishJob(&term.model);
+
+    const failed = try term.screen();
+    defer a.free(failed);
+    try std.testing.expect(std.mem.indexOf(u8, failed, "invalid_request_error") != null);
+    try std.testing.expect(std.mem.indexOf(u8, failed, "check /model and your API key") == null);
+
+    _ = term.typeText("continue");
+    const recovered = try term.screen();
+    defer a.free(recovered);
+    try std.testing.expect(std.mem.indexOf(u8, recovered, "continue") != null);
+    try std.testing.expect(std.mem.indexOf(u8, recovered, "›") != null);
 }
 
 test "assistant markdown shows lists, quotes, tasks, and rules" {
--- a/scripts/codex_ws_error_test.py
+++ b/scripts/codex_ws_error_test.py
@@ -1,0 +1,205 @@
+#!/usr/bin/env python3
+"""Issue #692 interactive Codex WS error-frame regression scenarios."""
+
+from __future__ import annotations
+
+import json
+import os
+from pathlib import Path
+
+from codex_ws_mock import CodexMock, RecordedRequest
+import codex_ws_test as base
+
+GENERIC_ERROR_CODE = "invalid_request_error"
+GENERIC_ERROR_SECRET = "raw-envelope-secret-must-not-leak"
+GENERIC_ERROR_TAIL = "generic-error-tail-must-be-truncated"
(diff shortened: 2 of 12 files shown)


justrach:
Folded into release/v0.0.282 (#696). Closing as superseded.

PR #694, opened 2026-08-30
fix(tui): recover poisoned TLS client generations

### LINKED ISSUE #691: tui/http: TlsInitializationFailed persists across turns until process restart
## Problem / failure mode

A transient network transition can put the launch-owned `std.http.Client` into a state where every later model request fails with `TlsInitializationFailed`, even after the user submits a new TUI prompt, waits, or changes network state. Quitting graff and resuming the same durable session in a new process recovers.

This is not limited to the line REPL. It affects the fullscreen TUI runtime's shared provider client and therefore ordinary root prompts, synthetic/idle-wake prompts, compaction/model calls, titles/recaps, and foreground/background subagents. Local-only trajectories such as `!cmd` are unaffected.

## Evidence

Two checked-in traces independently show launch-scoped persistence:

- `.graff/traces/a3334f40111960cd.jsonl:629-651`: after a successful long turn, the next request transitions through `WriteFailed` and `NetworkUnreachable` into `TlsInitializationFailed`; two additional user turns each spend all six retries on `TlsInitializationFailed` and return zero response bytes.
- `.graff/traces/78e55567de819be8.jsonl:2170-2236`: a foreground subagent and the root both exhaust TLS retries; later root prompts fail again, one intervening turn reports `UnknownHostName`, and TLS failures resume. The same process still fails after long idle gaps (including ~5 hours), ruling out a short-lived outage/backoff window.

The user reported the same sequence for runtime session `session-1788101995913-95134`: reconnecting to a better network and reprompting did not recover; quit + `/resume` did.

## Why this is a graff lifecycle bug

- One HTTP client is constructed for the launch and deinitialized only at process exit (`src/main.zig:295-303`).
- Fullscreen TUI stores that pointer in `ReplCtx`; every TUI model turn and child agent reuses it (`src/tui_launch.zig:80-97`, `src/repl_turn.zig:57-75`, `src/subagent_run.zig:269-284`).
- The six-attempt ladder retries the same client (`src/agent_request.zig:245-254`, `src/agent_request.zig:349-390`).
- Existing #177 poisoning is installed only *after* `client.request(...)` returns a Request (`src/http.zig:137-151`, `src/agent_stream.zig:112-129`). A TLS-initialization error raised during `client.request` happens before graff can mark a connection closing.
- CA prewarm errors are silently swallowed while the ready event is still released (`src/http_warm.zig:8-18`), so traces cannot distinguish a failed CA scan from a later TLS-client failure.
- TUI `/resume` resets Agent/WS state but retains the same launch-owned HTTP client; process replacement is the only current full reset.

This is adjacent to closed #177 but occurs earlier in request construction, outside that fix's cleanup scope.

## Expected behavior

After bounded retries exhaust on `TlsInitializationFailed`, a later TUI turn on a recovered network must be able to redial without restarting graff. Graff should also trace CA-prewarm failure and whether TLS failed during request construction, so the recovery path is diagnosable.

## Suggested approach / constraints

- Add a synchronized, launch-level HTTP client recovery path for request-construction TLS failures (or a safe fresh-client retry), rather than blindly replaying against the same client six times on every later turn.
- Do not deinit/swap the shared client while concurrent title, recap, compaction, or subagent requests can still be using it; client generation/ref-counting or a per-attempt fresh client may be safer.
- Preserve the existing post-request connection poison for send/read failures.
- Add a mock/TUI regression that fails TLS initialization for one client generation, then verifies a later ordinary prompt and a child trajectory recover without process exit.

yxlyx:
Closes #691.

## What changed

- Added a launch-level recovery manager for model HTTP traffic. Both non-streaming `http.postWatched` calls and streaming `Agent.postStreamWithClient` calls lease the active client generation.
- When `std.http.Client.request` returns `TlsInitializationFailed` before a `Request` exists, retire that generation and publish a CA-prewarmed replacement. Concurrent callers keep the retired generation alive until their final lease releases; stale reports do not rotate twice.
- Added a lifecycle gate around global recovery lookup/publication and teardown. Shutdown closes admission, drains active/retired leases and CA-ready waiters, then destroys owned clients; calls targeting the retired launch client are rejected while unrelated `std.http.Client` instances remain usable.
- Centralized the launch CA-ready barrier at the two request constructors, including direct `postWatched` users such as server compaction, and added distinct breadcrumbs for launch CA failure, replacement CA failure, TLS rotation, and final retry exhaustion.
- Corrected retry accounting so a six-attempt plan performs six attempts rather than seven, keeps one total-attempt budget when error classes change, and preserves 429/5xx attribution at exhaustion.
- Added deterministic regressions for real malformed TLS handshakes through both production catches, repeated and simultaneous stale-generation failures, replacement allocation/CA failures, shutdown races, lease reclamation, retry exhaustion followed by successful later root/child calls, TUI root-agent construction, foreground `runSub`, and background `execSubagent`/`agentOutput`.

## Why

A transient network transition can poison the launch-owned `std.http.Client` at request construction, before the existing post-construction connection cleanup has a `Request` to mark closing. Retrying the same client therefore leaves every later model-backed TUI trajectory failing until the process is replaced.

Generation leasing fixes that lifecycle boundary rather than masking individual failures. One poisoned client can be retired immediately while in-flight compaction, root, or child calls safely finish on the object they already leased, and later retries resolve the original launch pointer to the replacement.

The design intentionally applies to model request constructors rather than every best-effort launch HTTP consumer. The original launch client is never deinitialized by rotation because telemetry/fleet callers may still hold its pointer. The cost is one mutex-protected lookup per model request plus temporary retention of retired generations while calls remain in flight.

Reinitializing the shared client in place was rejected because concurrent users can still dereference it. Creating a fresh client for every request was rejected because it discards connection pooling and repeatedly reloads CA state. Relying on debug assertions or external job-join ordering was rejected because ReleaseFast teardown still needs an in-process lifetime guarantee.

## Verification

- Three independent five-agent adversarial review rounds covered streaming roots/synthetic wakes, foreground/background children and workflows, compaction/server-compaction, title/recap and CA observability, retry accounting, plus concurrency/memory/shutdown edges.
- `scripts/eval-tier1.sh` — green with the test-count ratchet updated to the verified suite size.
- Unit suite: **1,783 passed, 1 skipped**.
- Focused TLS suite: **54/54 passed**.
- Fullscreen TUI suite: **459/459 passed**.
- Real-binary TUI guards: **17/17 passed**.
- `zig build -Doptimize=ReleaseFast` — green.
- `git diff --check` — clean.


justrach:
Folded into release/v0.0.282 (#696), including the TLS-lifecycle follow-up. Closing as superseded. ADR is 0048.

PR #696, opened 2026-08-31
release/v0.0.282 — paste spans, Codex WS, TLS, session branches, teleport, DeepSeek lean

justrach:
Next cut after [v0.0.281](https://github.com/justrach/codegraff/pull/670) merged to main. No GitHub tag until asked. Download / notarization ids land after the build. Do not merge to main until asked.

## Folded from yxlyx

- **#674** / #673 — atomic `[Pasted text #N]` spans. Navigation and delete treat a chip as one unit; a typed lookalike stays literal.
- **#693** / #692 — Codex Responses `type:error` is a terminal API response. Bounded `last_api_error`; one-shot full-input rebuild on a stale `previous_response_id`.
- **#694** / #691 — lease/rotate the shared `std.http.Client` on request-construction `TlsInitializationFailed`. Follow-up hardens lease teardown (drain on shutdown, retry accounting, child coverage).
- **#697** / #689 — `--resume SOURCE --branch DEST` clone-on-write so two processes cannot clobber one `.session.json`. yxlyx reopened the same head as **#698** after #697 closed; already on this tip.
- **#679** — MCP HTTP/WSS reuse the warmed TLS client; MCP HTTP accepts gzip catalogs.
- **#695** — an empty tool catalog omits the `tools` field instead of serializing `"tools":,` (0.0.280 HTTP 400). Landed on the remote 282 tip while this cut was mid-fold; merged in.

## ADR remaps

`#694` / `#697` / `#679` numbered records 0042 / 0042 / 0043. Those slots are already TUI claims screen (`#666`) and Pi SWE (`#667`) on main.

| Source | Original | This cut |
| --- | --- | --- |
| TLS client generations (#694) | 0042 | **0048** |
| Session branches (#697) | 0042 | **0049** |
| MCP TLS reuse (#679) | 0043 | **0050** |

## Not folded

- **#675** DeepSeek flash thinking off + lean 15s bash — ADR 0048–0051 collide with this cut's 0048–0050. Leave on its own branch.
- **#677** sandbox teleport snapshots / GC leftover tars — leave on its own branch.
- **#277** Streamable HTTP OAuth / Smolify and **#200** idle localhost servers stay parked. Smolify is not a reserved core MCP. Inventory: [yxlyx-leftovers.md](https://github.com/justrach/codegraff/blob/release/v0.0.282/docs/yxlyx-leftovers.md).

## Closed superseded PRs

Folded heads closed after landing here: **#674 #693 #694 #679 #697**.

[...]

PR #697, opened 2026-08-31
Fix concurrent session resumes with clone-on-write branches

### LINKED ISSUE #689: Make /resume clone-on-write so concurrent TUI sessions can branch safely
## What changed / problem

Two independent terminal sessions cannot safely use one saved session as a common baseline and then continue as separate branches.

A real process-level reproduction against the scripted Codex backend exposed two related failure modes:

1. **Default line REPL `/resume` can lose the durable target.** A process loaded `baseline.session.json`, completed a turn with the baseline present in the provider request, and exited `0`. It printed:

   ```text
   resumed baseline.session.json — 2 message(s), gpt-5.6-sol via codex
   ...
   SEEN base=True turn=True
   ↩ session saved → _73
   line.session.json
   ```

   The completed turn was written to a corrupted `_73\nline.session.json`; `baseline.session.json` remained at two messages. This reproduces the user-visible “turn worked, then my changes were gone” behavior without a provider failure.

   The current unit test `typed session name survives readline reusing its line buffer` passes, but it drives `handleCommand` directly (`src/main_test.zig:204`) and does not run the full input → turn → autosave/final-save lifecycle. The apparent ownership fix is in `ownedSessionName` (`src/commands_misc.zig:148-157`) and the assignment after load is at `src/commands_misc.zig:501-550`; an integration test is needed because the real runtime still corrupts the target.

2. **There is no branch identity.** Two processes may load the same complete baseline while both remain alive, but both retain the same `root.session_name`. Whole snapshots are then written to the same `.graff/sessions/<name>.session.json`. The write lock (`src/session_lock.zig:1-17,68-90`) prevents a simultaneous clobber only for the duration of one write; it does not create ancestry, merge histories, or give each resumed process an independent tip. Sequential writes are still last-writer-wins.

ADR 0014 explicitly says that today one process owns a session file and leaves two-process inbox semantics unresolved (`docs/adr/0014-session-resume-carries-the-room-cursor.md:34-37`).

## Why this needs a runtime-level solution

Users reasonably treat `/resume SOURCE` like checking out a baseline: session A and session B should be able to continue different tasks without either process quitting, retargeting the other, or making the other branch disappear. A short file lock cannot provide that semantic; each continuation needs a distinct durable identity while retaining immutable ancestry back to `SOURCE`.

This also needs to cover **all terminal/TUI trajectories**, not only the default line REPL:

- `graff tui --resume SOURCE` and TTY `graff repl --resume SOURCE` load the root before launch, but fullscreen TUI creates a separate empty engine conversation (`src/tui_launch.zig:96-106`) rather than seeding it from `root.messages`.
- Fullscreen `/resume` is absent from `TUI/catalog.zig`/`TUI/dispatch.zig`; only startup `--resume` exists.
- TUI turn agents default their checkpoint target to `last` rather than the resumed root name (`src/agent.zig:166-170`, `src/repl_turn.zig:56-116`, `src/turn_checkpoint.zig:8-11`).
- The TUI dispatch returns early from `main` (`src/main.zig:487`) and bypasses normal `finalizeSession`/save-drain (`src/main.zig:528`, `src/session_run.zig:432-458`).
- The in-process ACP-backed fullscreen path must receive the same restored branch history; fixing only a rendered header would leave provider requests wrong.

The scripted/non-TTY `graff repl` path and `graff acp --resume` should either implement the same contract or explicitly reject unsupported resume/branch combinations. This issue is about TUI/runtime behavior, not GUI behavior.

## Proposed behavior

Introduce explicit clone-on-write session branching, for example one of:

```text
graff --resume SOURCE --branch BRANCH_A
graff tui --resume SOURCE --branch BRANCH_A
/resume SOURCE --branch BRANCH_A
```

or an equivalent `/fork BRANCH_A [SOURCE]` command.

The important contract is:

1. Read `SOURCE` as the common immutable baseline.
2. Mint or require a unique destination session identity for each continuation.
3. Copy/seed provider-native history, goal/todo state, compaction/cache metadata, peer cursor/inbox policy, and transcript generation into that destination.
4. Persist future turns, transcripts, checkpoints, and shutdown saves only to the destination branch.
5. Keep both processes usable concurrently; completion/exit of A must not terminate B or change B's save target.
6. Make source/parent ancestry visible in session metadata so `/sessions` can show `BRANCH_A ← SOURCE` and future merge/cherry-pick behavior remains possible.
7. Treat repository-file isolation separately and clearly: session-history branching must be safe by itself; `-w/--worktree` remains the mechanism for independent filesystem edits, and the UI should recommend/pair it when appropriate.

An automatic unique child name when a live owner already has `SOURCE` would be a useful convenience, but silent auto-forking should not make it unclear where work was saved.

## Acceptance tests

Add deterministic offline integration coverage using `scripts/codex_ws_mock.py`:

1. Seed `SOURCE` with a sentinel baseline turn.
2. Start A and B as separate real processes.
3. Have both resume/fork from `SOURCE`; wait until both report ready before sending either next prompt.
4. Send unique A/B sentinels concurrently and verify both provider requests contain SOURCE but not the sibling sentinel.
5. Keep A and B alive after their turns and reopen each destination from a third process; each must contain SOURCE + its own sentinel only.
6. Exit A first; prove B remains alive and can complete another turn. Exit B and reopen both branches again.
7. Assert no `SessionOpenInAnotherGraff`, corrupt/retargeted filename, lost turn, malformed transcript JSONL, or orphaned save queue.
8. Run equivalent host-wired coverage for fullscreen `graff tui`, TTY `graff repl`, startup compaction, `/new`, `/rewind`, tool checkpoints, and normal/active-turn shutdown. Use the headless TUI engine where possible and existing PTY guard infrastructure only for process/terminal lifecycle boundaries.
9. Add a regression test around the exact current line-REPL failure: `/resume baseline`, reuse the real input path for a subsequent prompt, complete the turn, EOF, and assert only `baseline.session.json` is updated with the turn.

## Constraints / trade-offs

- Do not merge concurrent whole-message arrays automatically; provider-native tool/reasoning blocks and compaction boundaries make a naïve append merge unsafe.
- Do not solve this with a process-lifetime exclusive lock on SOURCE; that only rejects the second user and does not provide branching.
- Preserve ADR 0014's no-room-replay guarantee. Define whether a branch copies the parent's inbox/cursors once or starts at room tail, then record that decision in a follow-up ADR because it is load-bearing.
- Session branching does not imply git worktree branching; expose the distinction rather than risking filesystem collisions.

## Environment / evidence

- Current repository HEAD and tracked `zig-out/bin/graff` on macOS.
- `zig build test --summary all`: **1752 passed, 1 skipped**.
- Duplicate search found no open/closed issue specifically covering same-baseline session branching or this `/resume` target corruption. Related historical work includes #289's short-duration session write lock and #114's git worktree isolation, neither of which provides conversation ancestry/branches.

yxlyx:
Closes #689.

## What changed

- Added `--resume SOURCE --branch DESTINATION` and `/resume SOURCE --branch DESTINATION` across the default line session, scripted/TTY REPL, fullscreen TUI, ACP startup, and one-shot runtime paths.
- Persisted branch ancestry, assigned each branch a fresh cache/session UUID, and atomically claimed new destination names so concurrent creators cannot both win.
- Seeded fullscreen TUI conversations from provider-native restored history and synchronized messages, goal/todo policy, strict/ultracode state, rename state, meters, and shutdown saves back to the selected durable branch.
- Added an offline process regression covering source immutability, two live branches, close-order independence, reopen behavior, input-buffer ownership, cache identity, transcript validity, and duplicate-destination races.
- Recorded the clone-on-write identity contract in ADR 0042.

### Problem / failure mode

Two processes resuming the same named session continued writing the same `.session.json`, so autosaves and shutdown writes could silently replace one continuation with the other. The fullscreen TUI also did not start from the restored provider-native conversation or reliably return its final state to the root session, which made branching only the line REPL insufficient.

### Reason for this approach

Clone-on-write gives each continuation an explicit durable identity while preserving the source as a stable common ancestor. A filesystem-exclusive destination claim closes the check-then-create race, and sharing one restore implementation across frontends keeps branch ownership, goal/todo state, compaction, transcripts, and final saves consistent.

Branches deliberately do not merge automatically: provider-native tool/reasoning envelopes and compaction boundaries cannot be safely combined by concatenating message arrays. Peer cursor/inbox state is copied once, while git worktree isolation remains a separate user choice. Existing plain `/resume SOURCE` keeps its same-tip compatibility behavior and therefore remains unsuitable for concurrent continuation.

### Rejected alternatives

A process-lifetime lock on the source would prevent corruption by rejecting the second user, but it would not support the requested concurrent work. Last-writer-wins merging was rejected because it preserves neither branch and can create invalid provider history.

## Verification

- Five independent TUI/runtime test agents exercised line REPL, scripted REPL, fullscreen TUI, ACP, one-shot, compaction, shutdown, metadata, malformed names, and concurrency edge cases.
- `python3 scripts/test-session-branching.py ./zig-out/bin/graff`
- Live fullscreen `/resume SOURCE --branch DEST` PTY regression
- Fullscreen startup/close-order/duplicate-race stress (8 race rounds)
- Fullscreen automatic and explicit compaction probes
- `scripts/eval-tier1.sh` (1,754 unit tests passed, 1 skipped; 460/460 TUI tests; all 17 PTY guard probes; build/reachability/SDK/invariants green)


justrach:
Folded into release/v0.0.282 (#696). ADR remapped 0042 → 0049 (0042 is TUI claims). Closing as superseded.

yxlyx:
Superseded because GitHub kept this PR pinned to the pre-fix SHA after its head branch was auto-deleted and recreated. Reopening against the current conflict-free head.

PR #699, opened 2026-08-31
fix(tui): harden split input and paste recovery

### LINKED ISSUE #537: TUI escape/death audit round 2: 12 unverified findings
Triage list from the post-v0.0.262 escape/death audit: 12 findings deduped below the verification cap. Several are cross-lane reproductions of the confirmed issues (link when triaging); the distinct ones (C0 controls in the stream tail, the `tui: render error` literal-free, kitty modifier latch after `.drop`, control bytes inside bracketed paste quitting the TUI) need their own verification pass.

- [ ] **[high]** (grok-build-reference) `!cmd` and the `@` file picker shell out synchronously inside the read loop, freezing the whole TUI for up to 20s
- [ ] **[high]** (pty-fuzz) A 50 ms gap between an escape sequence's ESC and its body fires a phantom Escape and types the whole body into the composer
- [ ] **[high]** (pty-fuzz) A bracketed paste whose CSI 201~ terminator is split at the ESC wedges the composer in paste mode permanently
- [ ] **[high]** (pty-fuzz) Control bytes inside a bracketed paste bypass the paste guard and quit the TUI with no confirmation
- [ ] **[high]** (pty-fuzz) escapeSeq/ss3 scan past an embedded ESC, so one truncated mouse event makes the NEXT event's body appear as typed text
- [ ] **[high]** (split-model) A read boundary inside CSI 200~ loses paste_start, so the pasted body is executed as key bindings (newlines submit turns, 0x03 quits the TUI)
- [ ] **[high]** (split-model) A lost paste_end latches Model.pasting and key.zig in_paste forever — Enter never submits again and there is no recovery path
- [ ] **[high]** (split-model) After the escape_key stall verdict, the rest of the sequence is typed verbatim into the composer — run.zig never arms orphan recovery on that branch
- [ ] **[medium]** (escape-residue) Live stream tail strips ESC but not C0 controls, so CR/BEL/BS from the model are painted straight to the terminal (blanked rows, repeating beeps)
- [ ] **[medium]** (death-paths) The 'tui: render error' fallback frees a string literal every frame — the render-failure path is itself fatal
- [ ] **[medium]** (grok-build-reference) Fatal-signal restore misses SEGV/BUS/ILL/FPE, so a release-build crash strands the terminal — and the ABRT path erases the panic message
- [ ] **[medium]** (split-model) The .drop verdict clears the byte buffer but not key.zig's latched modifier state, so a lost kitty release turns the next bare Backspace into delete-to-start

Refuted during verification (do not re-file without new evidence):
- Headless SGR mouse debris mid-read-buffer — code observation accurate but the trigger is unreachable as described.

yxlyx:
## What changed

- Preserve bounded terminal-sequence framing across split reads, including CSI, SS3, X10/SGR mouse, kitty CSI-u, OSC, and bracketed-paste markers.
- Route paste input before global controls so pasted C0 bytes, mouse reports, and modifier sequences cannot quit, cancel, navigate, or mutate held-key state.
- Clear stale modifiers on every abandoned-sequence path and make normal-turn/background thread-spawn failures report asynchronously instead of blocking the TUI loop.
- Serialize `/model` mutation while background compaction owns the shared runtime context.
- Add reached parser/simulator/runtime tests and an offline trajectory-proven PTY probe across idle, live-turn, compact, bash, and file-list trajectories.

### Problem / failure mode

The audit findings in #537 shared one state-machine failure: when terminal input was split, delayed, truncated, or filled the read buffer, the TUI could lose framing and reinterpret protocol bytes as user controls. That could cancel live work, quit during paste, erase a draft through stale Super state, fabricate wheel/click events, type protocol debris, or freeze the renderer if thread creation failed.

### Reason for this approach

Recovery now distinguishes strong evidence (a dropped non-lone sequence head) from an ambiguous genuine Escape. Strongly evidenced heads retain exact bounded framing; mismatches fail closed and fresh human input wins. Paste markers restore paste state before same-read payload, while all runtime trajectories share the same guarded batch dispatcher. This fixes the root state transitions rather than adding per-sequence string filters.

- A lone Escape waits about 300 ms while idle and up to one second while cancellable work is live, because it is indistinguishable from the first byte of a delayed paste marker. Ctrl-C and kitty CSI-u Escape remain immediate.
- Carry-expired ambiguous short CSI/SS3 tails become text; self-identifying framed protocols can still recover within the bounded window. Preserving human text is safer than guessing an arrow from `[Alice]`.
- Proven but oversized/misaligned mouse debris is dropped rather than fabricated into an action.
- The test-count floor is 1860 because macOS deterministically reports the remaining platform test as one skip; non-skipped platforms still run 1861.

### Rejected alternatives

- Calling the global input reset on every drop would destroy paste/orphan recovery along with stale modifiers.
- Broad late-tail guessing corrupted ordinary bracketed prose and fabricated mouse fields when a prefix was missing.
- Running provider/tool work inline after `Thread.spawn` failure preserved completion at the cost of freezing all input and rendering.

## Verification

- `scripts/eval-tier1.sh`: green
- main suite: 1860 passed, 1 platform skip
- TUI suite: 505/505 passed
- PTY/tuiguard: 18/18 passed, including `scripts/test-tui-escape-split.py`
- formatting, 600-line ceiling, test reachability, invariants, and SDK drift: green

Closes #537.


justrach:
Folded onto `release/v0.0.283` (PR #703) as `edfb065` + `0823491`.

Kept 283's #704 tuiguard deadlines (ThreadPool + killpg) and the 1868 suite floor; added `test-tui-escape-split.py` as the 18th probe with a 90s budget. Did not take the 1860 macOS floor drop.

Tier 1 green on the push (TUI 505/505, tuiguard 18/18). #537 closed from the release branch. This PR can close as superseded once you are happy — leaving that to you.

justrach:
Shipped in **v0.0.283** via #703 (`edfb065` / `0823491`). Closing as superseded — not merging this branch to `main`.

PR #705, opened 2026-08-31
fix(credentials): complete key-store durability follow-ups

### LINKED ISSUE #405: credential_store follow-ups: symlink semantics, keys_cli coverage, dir fsync
Deferred minors from the review of fix/oauth-atomic-writes-perms (merged for v0.0.239).

1. **Rename-into-place replaces a symlinked target instead of writing through it.** `credential_store.replaceFile` on a `settings.json -> real.json` symlink replaces the symlink with a regular file, leaving `real.json` stale. Decide the policy (resolve symlinks before staging, or document replacement) — today it is silent.
2. **`keys_cli.zig` conversion is untested on its live platforms.** The file path is dead on macOS (`storeKey` short-circuits to the keychain), so the atomic conversion there runs only on Linux/Windows where we have no local coverage. Also the comment overstates 0600 on Windows: `private_file` degrades to `.default_file` and the chmod block is comptime-elided (`has_executable_bit == false`).
3. **Directory fsync (power-loss durability).** `replaceFile` fsyncs the temp file but not the containing directory, so the rename itself is not durable across power loss (fine across process crashes). Matches the pre-existing in-tree pattern (learn_store, eval_memory); qualify or harden as a batch if it ever matters.

yxlyx:
Closes #405 after the symlink-policy portion landed separately in #685.

## What changed

- Sync the exact destination directory retained by `createFileAtomic` after an atomic credential-file replacement on POSIX.
- Exercise the real non-macOS file-backed key store through merge/load, repeated read-modify-write, restrictive-mode, nested-target, and temp-file cleanup cases.
- Describe the actual platform security model consistently: macOS Keychain, POSIX mode 0600, and inherited home-directory ACLs on Windows.

### Problem / failure mode

Atomic replacement synced the new file contents but not the directory entry, so a hard power loss could lose the rename. Reopening the containing directory from the destination path after replacement would add a race: a symlinked parent could be retargeted between rename and fsync, causing the wrong directory to be synced or reporting failure after the replacement already succeeded. The Linux/Windows key-store path also lacked live-platform test coverage, and user-facing help overstated Windows as providing POSIX 0600 semantics.

### Reason for this approach

`createFileAtomic` already retains the exact directory handle used for its temporary file and rename. Syncing `atomic.dir` after `atomic.replace` pins the actual destination directory without a second path traversal. Tests use the real file-backed `storeKey` path on its live platforms and verify preservation and permissions rather than duplicating serialization logic in a mock.

Windows keeps its existing weaker durability/security model: directory `FlushFileBuffers` is not portable and Zig file modes do not express Windows ACLs, so the implementation relies on atomic replacement, synced file contents, and the containing home-directory ACL. The docs now say this explicitly. Multi-process read-modify-write locking is a separate pre-existing concern and is outside #405's deferred items.

### Rejected alternatives

- Reopening `dirname(dest)` after rename was rejected because pathname resolution can race with symlink-parent retargeting.
- Duplicating the symlink policy was rejected because #685 already implemented and merged item 1.
- Claiming 0600 on Windows was rejected because `.default_file` carries no equivalent ACL guarantee.

## Verification

- `scripts/eval-tier1.sh`: green (`1865` passed + 1 platform skip, `465/465` TUI, `17/17` PTY).
- Final pre-push rerun: fmt, line ceiling, spec, reachability, build, `1865` + 1 skip, `465/465` TUI, invariants, and SDK all passed; one unrelated intermittent `test-tui-resize-anchor.py` probe missed a repaint once and then passed immediately via `python3 scripts/test-tui-resize-anchor.py zig-out/bin/graff`.
- Independent focused review validated `atomic.dir` lifetime and the path-race fix; no in-scope blockers remain.

### DIFF
--- a/src/cli.zig
+++ b/src/cli.zig
@@ -251,7 +251,7 @@ pub const usage_text =
     \\  graff login codex [--refresh]    ChatGPT/Codex OAuth login (PKCE)
     \\  graff login kimi                 Kimi Code OAuth login (device-code)
     \\  graff login xai                  Grok/SuperGrok OAuth login (device-code)
-    \\  graff key set <provider> <key>   store a key (macOS Keychain, else 0600 file)
+    \\  graff key set <provider> <key>   store a key (Keychain; POSIX 0600 file; Windows home ACL)
     \\  graff key list                   show which providers have keys
     \\  graff models [refresh]           list the live catalog; refresh Codex + models.dev metadata
     \\  graff route <model>…             dry-run which provider/billing a model lands on (no API call)
--- a/src/credential_store.zig
+++ b/src/credential_store.zig
@@ -7,11 +7,10 @@
 //! silent logout with nothing left to recover from. Write a per-writer-unique
 //! temp file in the target's own directory, fsync it, then rename over the
 //! target, so a reader sees either the whole old file or the whole new one.
-//!
-//! The guarantee is against a crashed process, not against power loss: the temp
-//! file is fsynced but the containing directory is not, so some filesystems can
-//! still lose the rename across a hard power cut. Same shape — and same limit —
-//! as learn_store.writeAtomicReplace / eval_memory.writeNotes.
+//! On POSIX, fsync the containing directory after the rename so that directory
+//! entry is also durable across power loss. Zig 0.17 cannot portably flush
+//! Windows directory handles, so Windows keeps the atomic replacement and
+//! synced file contents but has a weaker power-loss guarantee.
 
 const std = @import("std");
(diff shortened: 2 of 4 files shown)


justrach:
Folded onto `release/v0.0.283` as `fe158a8` (PR #703). #405 is closed. This PR can be closed as superseded once you are happy the fold is intact.

justrach:
Shipped in **v0.0.283** via #703 (`fe158a8`). Closing as superseded — not merging this branch to `main`.

PR #707, opened 2026-08-31
fix(tui): preserve legacy Alt navigation

### LINKED ISSUE #524: TUI audit triage: 16 unverified findings from the out-of-the-box review
The same TUI out-of-the-box audit that produced the 8 confirmed issues also surfaced 16 findings that were deduped but not adversarially verified (capped at top 8 by severity). Each needs a verification pass before being trusted; refute or promote to its own issue.

- [ ] **[high]** Composer cursor math is byte-wise, so Backspace/←/→ split UTF-8 codepoints and corrupt both the buffer and the rendered prompt (`TUI/input.zig`)
- [ ] **[high]** Bracketed-paste bytes are re-decoded as keystrokes, so pasted tabs are silently deleted and CRLF lines are doubled (`TUI/keys.zig`)
- [ ] **[medium]** ESC ESC <CSI> emits a real Escape, so Alt+Left/Right/Up cancels the running turn (`TUI/key.zig`)
- [ ] **[medium]** Stale super/alt latch: unmodified kitty keys never resync `held`, so Backspace becomes delete_to_start and 'z' becomes undo (`TUI/key.zig`)
- [ ] **[medium]** Composer box borders are 2 columns narrower than its body rows, so the corners never meet the sides (`TUI/chrome.zig`)
- [ ] **[medium]** Image preview card is counted as 4 rows but painted as 3: frame is one row short of the terminal and prompt_origin points one row below the real composer, so hovering the chip closes the card (`TUI/render.zig`)
- [ ] **[medium]** Diff repaint never invalidates on a width change: widening the terminal leaves unchanged rows painted only to the old width, with a raw default-background band on the right (`TUI/run.zig`)
- [ ] **[medium]** Collapsed tool summary counts the ⚙ start row and the ✓ done row as two separate tools, so "Called N tools" is double the real count and increments as a single tool finishes (`TUI/scrollback.zig`)
- [ ] **[medium]** Scroll offset is anchored to the bottom of the transcript, so a scrolled-up viewport drifts downward line-for-line as the live turn appends rows (`TUI/render.zig`)
- [ ] **[medium]** Any assistant answer line beginning with ✓/⚙/✗/⊘ is harvested into the transcript as a phantom tool entry and then repeated inside the reply (`TUI/turn.zig`)
- [ ] **[medium]** /compact runs the summarization API call on the UI thread — the whole TUI freezes with no indication (`TUI/dispatch.zig`)
- [ ] **[medium]** Quitting during a live turn never cancels it: the terminal is restored, then deinit blocks joining a thread that keeps running tools (`TUI/app.zig`)
- [ ] **[medium]** Ctrl+P palette: render highlights item (sel % 12) while Enter runs item min(sel, n-1); 22 of 34 commands are unreachable (`TUI/chrome.zig`)
- [ ] **[medium]** /history is a self-recalling no-op: applyLine records the slash command itself before dispatching, so recallPrev always returns "/history" (`TUI/dispatch.zig`)
- [ ] **[medium]** Typing `@` mid-sentence hijacks the keyboard into the file picker; the characters typed after it never reach the composer and are destroyed on Esc (`TUI/keys.zig`)
- [ ] **[medium]** Queued steer text is stranded (and later resurrected in the wrong session) because run.zig drains exactly one queue entry, and only when a turn completes (`TUI/run.zig`)

yxlyx:
Closes #524. This PR is intentionally stacked on #699 and contains one additional commit; once #699 merges, this PR can be retargeted to `main` without changing its diff.

## What changed

- Decode legacy `ESC ESC <CSI>` Alt+Left/Right/Up as one navigation event instead of Escape followed by navigation.
- Preserve framing across every read split, bounded delayed delivery, full-buffer recovery, and dropped-head recovery.
- Keep genuine lone/double Escape behavior after the ambiguity window and preserve recognizable bracketed prose instead of guessing it is navigation.
- Add reachable fullscreen-TUI regressions for idle, live turn, background compact/bash/files, overlay, and bracketed-paste trajectories.

### Problem / failure mode

Some terminals encode legacy Alt+Arrow as `ESC ESC` plus CSI. The parser previously emitted the first Escape immediately, so the shared TUI dispatcher could cancel a model turn or background operation, close an overlay, end paste, or arm rewind before processing the arrow. Split and delayed terminal reads made this destructive behavior timing-dependent.

### Reason for this approach

The incorrect event originates in the shared input framing path, so the fix keeps the complete legacy sequence ambiguous until it can be decoded once. The bounded stall/recovery state from #699 already tracks live operations and late terminal tails; building on it protects every fullscreen TUI trajectory without adding downstream guards or duplicating #537's broader recovery machinery.

The terminal byte stream cannot perfectly distinguish `ESC ESC [Ax]` as Alt+Up then `x]` from genuine double Escape followed by literal `[Ax]`. The implementation uses narrow evidence: exact navigation, split input, and navigation followed by ordinary coalesced text win as navigation; recognizable bracketed prose wins as text; an incomplete sequence beyond the existing ~300 ms idle / ~1 s live window resolves as genuine Escape. This preserves cancellation responsiveness while covering the reported 50–250 ms delivery range.

### Rejected alternatives

- Suppressing Escape only in live-turn code was rejected because the same phantom event affects background operations, overlays, paste, rewind, and idle navigation.
- Copying the old #524 draft was rejected because it mixed paste normalization, stale modifiers, SS3/orphan recovery, and other #537 concerns.
- Basing directly on `main` was rejected because the delayed runtime seam needs #699's operation-aware stall/recovery behavior; restating that work would create overlapping PRs.

## Verification

- `GRAFF_TUIGUARD_JOBS=1 scripts/eval-tier1.sh` — green in 562s.
- Unit suite: `1860` passed + 1 platform skip (`1861` total).
- TUI suite: `514/514`.
- PTY/tuiguard: `18/18`.
- Formatting, 600-line ceiling, test reachability, build, invariants, and SDK drift checks all passed.
- Independent implementation, adversarial-test, and final-review agents exercised split/delayed input, coalesced printable keys, `[Alice]` prose, unmodified and `1;3` CSI forms, all background operation kinds, overlay movement, paste, full-buffer recovery, and genuine Escape behavior. Final review reported no blockers.

### DIFF
--- a/TUI/issue_524_tests.zig
+++ b/TUI/issue_524_tests.zig
@@ -1,0 +1,239 @@
+//! Reachable fullscreen TUI regressions for legacy ESC ESC CSI navigation.
+
+const std = @import("std");
+
+const app = @import("app.zig");
+const engine = @import("engine.zig");
+const key = @import("key.zig");
+const recover = @import("key_recover.zig");
+const stall = @import("run_stall.zig");
+const Term = @import("sim.zig").Term;
+
+const Trajectory = enum { idle, live_turn, compact, bash, files };
+const trajectories = [_]Trajectory{ .idle, .live_turn, .compact, .bash, .files };
+
+const Fixture = struct {
+    term: Term,
+    bg: ?*engine.BgOp = null,
+
+    fn init(trajectory: Trajectory) !Fixture {
+        var fixture: Fixture = undefined;
+        fixture.bg = null;
+        fixture.term.init(std.testing.allocator, 80, 24);
+        errdefer fixture.term.deinit();
+        try fixture.term.model.push(.user, "existing turn");
+        switch (trajectory) {
+            .idle => {},
+            .live_turn => {
+                const job = try std.testing.allocator.create(engine.Job);
+                job.* = .{ .gpa = std.testing.allocator, .history = &.{}, .params = .{}, .stream = .{}, .threaded = false };
+                try fixture.term.model.push(.pending, "");
+                fixture.term.model.pending = job;
+            },
+            .compact, .bash, .files => {
+                const op = try std.testing.allocator.create(engine.BgOp);
+                op.* = .{ .kind = switch (trajectory) {
+                    .compact => .compact,
+                    .bash => .bash,
+                    .files => .files,
+                    else => unreachable,
+                }, .gpa = std.testing.allocator, .threaded = false };
+                try fixture.term.model.push(.pending, "background operation");
+                fixture.term.model.bg = op;
+                fixture.bg = op;
+            },
+        }
+        return fixture;
+    }
+
+    fn deinit(self: *Fixture) void {
+        if (self.bg) |op| {
+            self.term.model.bg = null;
+            std.testing.allocator.destroy(op);
+        }
+        if (self.term.model.pending) |job| {
+            self.term.model.pending = null;
+            std.testing.allocator.destroy(job);
+        }
+        self.term.deinit();
+    }
+};
+
+fn expectAlive(fixture: *Fixture, trajectory: Trajectory) !void {
+    try std.testing.expect(!fixture.term.model.cancel_requested);
+    try std.testing.expect(fixture.term.last_effect == .stay);
+    if (trajectory == .live_turn) try std.testing.expect(fixture.term.model.pending != null);
+    if (trajectory != .idle and trajectory != .live_turn)
+        try std.testing.expect(fixture.bg != null and !fixture.bg.?.cancelled);
+}
+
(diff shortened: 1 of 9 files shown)


justrach:
Folded onto `release/v0.0.283` (PR #703) as `e674794` — this is the complete CSI-only Alt+arrow framing (`ESC ESC [` stays one nav event; SS3 `ESC ESC OA` stays Escape). Replaces the smaller leftover that was already on the cut.

Leaving this PR open for you to close as superseded.

justrach:
Shipped in **v0.0.283** via #703 (`e674794`). Closing as superseded — not merging this branch to `main`.

PR #764, opened 2026-09-06
feat: remote control over the account relay (#722)

### LINKED ISSUE #722: Account-scoped session control when logged into Codegraff
## What I want

If I am logged into Codegraff (`graff login` / the site account), I should have one place that lists **my** sessions and lets me control them — not only the ones in the current cwd, and not only from the machine that started them.

Today I can `/sessions` and `/resume` on a device (ADR 0059), `/tell` / `/peek` a live peer on that device (#469), and hit `graff serve` with a shared token. None of those is “I signed in, here are my sessions.”

## Control surface (v1)

When the same Codegraff login is present:

- **List** live and saved sessions I own (title, cwd/origin, model, last activity, running or idle).
- **Peek** a live one (what it is doing now) without attaching.
- **Steer** / queue a prompt into a live one.
- **Cancel / pause** a running turn.
- **Resume** a saved one on this device (tools stay in the current cwd, same as ADR 0059).

CLI (`graff sessions`, `/sessions` when logged in) and a site/native page that uses the same login are both fine. One API, two skins.

## What already exists (do not rebuild)

- Device-local presence + `/tell` / `/peek` (#469, closed). That issue explicitly deferred **cloud relay** and the auth/trust model.
- Device-wide saved-session discovery (#712 / ADR 0059). Local files only.
- TUI `/resume` picker is a separate local UX hole (#719).
- `graff serve` `POST/DELETE /v1/sessions` is token-auth for `--json` children, not the Codegraff account.
- ACP `session/load` is still unimplemented.

## Non-goals

- Not a second agent runtime. The engine stays a subprocess / in-process ACP client; this is a control plane on top of sessions we already save and announce.
- Not edit locking or CRDT merge (#469).
- Not replacing `graff login` with ACP `authenticate`.

## Why now

I start sessions from the TUI, the line REPL, and the native app. If I am logged in, I should be able to see and steer them from wherever that login is valid, instead of SSH-ing back to the box that happened to run `/resume`.

justrach:
Closes #722 (v1).

`graff remote-control` is `graff serve` with no listener: it registers with the gateway using the `graff login` key, long-polls the account's relay for commands, and streams each session's event tape back as it lands. Nothing on the machine is bound.

`graff remote` lists the machines and drives their sessions (`new`, `send`, `answer`, `tail`, `cancel`, `close`) with the same NDJSON a serve client gets. The machine keeps the complete tape; the relay keeps a recent window, and a viewer that falls behind gets a replay from the machine.

- `serve_create.createFromBody`: the transport-free half of session creation, shared by both bridges.
- `serve` exposes its child helpers (`writeChildLine`, answer-state, drop/reap) for the remote drain.
- Gated on the `sessions` key scope, so a plain `api` key can never reach a machine.
- `docs/remote-control.md`, CHANGELOG entry, 7 unit tests (all compiled in per tier 1).

Verified end to end against a local relay and then the production relay, including from a second machine (`scripts/e2e-remote-control.py`: a cloud sandbox lists this machine, creates a session on it, and the session's `hostname` names this machine, not the sandbox). Also: create, turn, list, close, resume with history intact, control-request errors surfaced, and gap → reattach → replay after a relay restart.

Security audit (docs/remote-control-security.md): remote control now needs the dedicated `remote` key scope rather than `sessions`; a viewer cannot start an unattended (`yolo`) session unless the machine was started with `--yolo`; a refused key stops the machine instead of retrying forever. The upload half moved to `remote_upload.zig`. Verified locally and against the production relay (403 without the scope, 403 on a `yolo` request against a non-yolo machine, normal turns unaffected).

Follow-ups noted on #722: shareable per-session tokens, sessions this supervisor did not start, streaming request responses, ACP `session/list` / `session/load` over relayed sessions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,11 @@ current is part of cutting a release.
   their sessions; `graff remote new` / `send` / `answer` / `tail` / `cancel`
   / `close` drive them, with the same NDJSON events a serve client gets. The
   machine keeps the complete tape; the relay keeps a recent window and a
-  viewer that falls behind gets a replay from the machine. Gated on the
-  `sessions` key scope, so a plain `api` key can never reach a machine.
-  `docs/remote-control.md`.
+  viewer that falls behind gets a replay from the machine. Gated on the new
+  `remote` key scope, so neither a plain `api` key nor a history-only
+  `sessions` key can reach a machine, and a viewer cannot start an
+  unattended session unless the machine was started with `--yolo`.
+  `docs/remote-control.md`, `docs/remote-control-security.md`.
 - Prompt-cache affinity is the git root, or one scratch seed when there is
   no `.git`. Eval sandboxes and worktrees can reuse a warm system+tools
   prefix instead of paying it on every leaf cwd. An offline test fails if
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,16 @@ current is part of cutting a release.
 
 ## Unreleased
 
+- `#722`: remote control. `graff remote-control` is `graff serve` with no
+  listener: it dials out to your Codegraff account and takes commands from
+  the account's relay, so nothing on the machine is ever bound. From anywhere
+  the same `graff login` is valid, `graff remote` lists the machines and
+  their sessions; `graff remote new` / `send` / `answer` / `tail` / `cancel`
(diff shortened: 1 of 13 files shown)


justrach:
Superseded by a rebase onto `release/v0.0.295`. Closing this copy to keep one PR.

PR #767, opened 2026-09-06
Breadth-first Project layout, list_dir near-miss hints, redirects are not constraints

justrach:
## What changed

- **Project layout is selected breadth-first.** `repo_map` now picks every top-level entry before any child, then each directory's first child before any second, under the same 300-entry / 6 KB caps. Emission is still a depth-first path list. The trailer names the top-level directories that were not fully expanded.
- **`codedb list_dir` on a missing path points at its neighbours.** New `list_dir_nearmiss.zig` appends the closest sibling entries (edit distance over a case- and separator-insensitive form; containment counts as close) and the parent's listing to the not-found error.
- **`note_constraint` records vetoes of a way of working, not redirects.** The root prompt sentence, its byte-pinned golden, and the tool description now say a course correction of this turn's target is followed, not recorded.
- ADR 0068 records the three decisions. A tier-2 case (`list-dir-near-miss-hint`) guards the hint. `test_count_baseline` is raised to 1988, as the ratchet check asked.

## Why

The root prompt tells the model the Project layout "is the tree" and to read from it instead of spending turns on `ls`. Filled depth-first, a workspace with about sixty top-level entries showed eight of them, so a folder the user named with a one-letter slip looked absent. The model then treated the name as a remote host, got a bare "was not found" from `list_dir` with nothing to try next, recorded the misreading as a standing constraint, and asked the user where the folder was. A plain listing six calls later answered it.

Three harness behaviours compounded, so all three change together: a tree that omits top-level names while claiming completeness, a preferred tool whose failure reply is a dead end, and a constraint trigger worded so that any "no, ..." reads as a veto.

Trade-offs: a fat directory shows fewer of its own files in the layout (siblings share the deep budget; `list_dir` still lists any of them on demand), and the not-found reply grows by up to about 1.3 KB of sibling names. Rejected: advertising codedb's `file <fuzzy-name>` instead of the hint (an extra tool hop the model has to think of; the hint arrives on the call it already made).

## Verification

- `zig build` + `zig build test` on the CI-pinned zig and on 0.16.0: 1988 passed, 1 skipped, 0 failed.
- Tier 1 green (fmt, lines, spec, reach, build, tests, tui, tuiguard, invariants, sdk).
- Tier 2 `--only list-dir-near-miss-hint` passes.
- The new layout rendered on a ~60-entry workspace shows every top-level entry inside the 6 KB cap; the old fill showed eight.


justrach:
Navigation changes integrated into release/v0.0.295 as 07356fe6, with ignore-aware sibling hints and bounded layout selection. The older constraint-policy changes were omitted in favor of #799. Release checks and the directory-hint behavior case passed.

PR #774, opened 2026-09-07
Add /version for running binary status

### LINKED ISSUE #769: Add /version with running and latest release status
A long-running terminal session keeps executing the version that launched it even after `graff update` replaces the installed binary. `/new` and `/resume` change conversation state inside that same process, so they do not make it obvious which version is currently running or whether it matches the latest release.

Add `/version` to the terminal command surfaces. It should:

- show the version compiled into the running process;
- report whether that version is current, older than the latest release, newer/dev, or unable to check;
- reuse the same release/version comparison as `graff update --check`;
- when an update has been installed, clarify that quitting and starting `graff` again is required, while `/new` and `/resume` do not reload the executable.

yxlyx:
## What changed

- add `/version` to the terminal command surfaces
- report the version compiled into the running process and its latest-release status
- share release classification with `graff update --check`
- keep the fullscreen release lookup off the render/input thread

## Why

A long-running session keeps the executable loaded at launch, so `/new` and `/resume` can make an installed update look active when it is not. Sharing one comparison policy prevents the updater and terminal status from drifting, while the background fullscreen check avoids blocking terminal interaction.

Closes #769

## Verification

- `scripts/eval-tier1.sh`

### DIFF
--- a/TUI/bgop.zig
+++ b/TUI/bgop.zig
@@ -1,4 +1,4 @@
-//! Background engine ops. `/compact`, `!cmd` and the @-file list used to call
+//! Background engine ops. `/compact`, `/version`, `!cmd` and the @-file list used to call
 //! straight into the engine from keys.handle, i.e. from inside run.zig's byte
 //! parse loop: no render, no poll, no key handling, no cancel for the whole
 //! round trip — a 5-60 s freeze indistinguishable from a hang (#533).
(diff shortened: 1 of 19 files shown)


justrach:
Superseded on the release branch: release/v0.0.293 already ships a superset of this (running-binary /version status plus /update for next-launch install). Nothing left to merge here.

PR #775, opened 2026-09-07
Fix Project Settings switch submission

### LINKED ISSUE #772: Project Settings switches close the dialog and submit stale values
## What happens

In the native GUI, open the project switcher and choose **Project settings**. Clicking either **Auto-approve tools** or **Start MCP servers** immediately closes the settings dialog.

The click also acts like **Save** for unrelated pending Name or Default model edits, but the clicked switch's new value is not persisted when the dialog is reopened.

## Expected

Switch clicks should update only the dialog's draft state and keep Project Settings open. **Save** should persist the complete draft, while **Cancel** should discard it.

## Cause and fix

The shared switch renders a `<button>` without `type="button"` inside the settings `<form>`, so the browser treats the switch as a submit control. React's batched update then submits the previous switch value.

Set the switch control to `type="button"` and add a regression test covering both switches, dialog visibility, draft persistence, Save, and Cancel.

yxlyx:
## What changed

- Mark the shared switch control as `type="button"` so it cannot implicitly submit a surrounding form.
- Extend the Electron projects flow to verify both Project Settings switches keep the dialog open, Cancel discards the draft, and Save persists the complete draft.

## Why

Project Settings is a form. Without an explicit button type, both switches were treated as submit controls: clicking one closed the dialog, saved unrelated pending Name or Default model edits, and submitted the previous switch value.

Fixing the shared switch primitive gives it the correct semantics anywhere it is embedded in a form. The settings form remains intentionally submittable through its normal submission path and explicit Save button. Handling this only in Project Settings was rejected because it would leave the reusable switch unsafe in other forms.

## Verification

- `npm run build`
- `npm run test:desktop` — 126 passed
- Electron Project Settings regression assertions passed in the projects flow
- `scripts/eval-tier1.sh`
- Pre-push tier 1

Closes #772


justrach:

justrach:
Accepted: merged into release/v0.0.293 as f0f7a24b (branch tip c010765f), verified with the full unit + TUI suites and tier-1 checks.

PR #779, opened 2026-09-07
Native folder picker: trailing slash and fuzzy rank

### LINKED ISSUE #776: Native GUI: folder picker should normalize directory paths and rank fuzzy matches
## Symptom

The native **Open a folder** picker treats its field only as exact-path navigation. Editing the field clears the folder results until a valid full path is submitted, and the visible directory entries remain alphabetically ordered. A user who knows a folder's approximate name but not its exact path cannot discover or prioritize likely matches.

Resolved directory paths are also displayed without a trailing `/` (for example, `/Users/example` instead of `/Users/example/`). This makes it less clear that the value is a directory and less convenient to continue typing a child path.

## Cause

The picker fetches one exact directory at a time, clears that listing on every edit, and sorts only the directory's immediate children by name. It has no query-aware candidate ranking or directory-path formatting.

## Expected fix

- Display resolved directory paths with a trailing separator, while handling `/` correctly.
- As the user types a partial path or folder name, show candidate folders ranked by match quality, with the closest matches first, following the fuzzy-picker behavior used by `/model`.
- Preserve exact-path navigation and folder selection.

justrach:
Closes #776.

The Open a folder picker treated the path field as exact navigation only: editing it cleared the listing, children stayed alphabetical, and a resolved directory showed without a trailing `/`.

This keeps the current folder visible while you type a partial path or name, ranks candidates with the same fuzzy score as `/model` (closest first), and displays resolved directories with a single trailing `/` (root stays `/`). Go, Open folder, and row selection still require an exact path.

**Tested**
- `node --experimental-strip-types --test` on `lib/fuzzy.test.ts` and `lib/folder-picker.test.ts` (path display, `//` join, query split, ranking)
- Native package test script for the new files plus existing workspace helpers
- `scripts/eval-tier1.sh` via pre-push (green)
- Desktop visual fixture (`project-recovery-visual.cjs`) now asserts the trailing slash and that a typed name ranks the closest folder first; the existing invalid-path / late-listing checks stay in place. The Electron visual runner was not available in this environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ current is part of cutting a release.
 
 ## Unreleased
 
+- `#776`: the native Open a folder picker keeps listing candidates while you
+  type, ranks them like `/model` (closest name first), and shows a resolved
+  directory with a trailing `/` (never `//`). Exact-path Go / Open folder
+  and row selection are unchanged.
 - Prompt-cache affinity is the git root, or one scratch seed when there is
   no `.git`. Eval sandboxes and worktrees can reuse a warm system+tools
   prefix instead of paying it on every leaf cwd. An offline test fails if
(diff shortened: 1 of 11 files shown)


justrach:

PR #781, opened 2026-09-07
Add /update to install a release for the next launch

### LINKED ISSUE #773: Add /update to install for the next launch without interrupting the current session
## Problem

Updating should not require opening another terminal or interrupting a conversation. The automatic restart and launch-flag replay proposed in #770 are not the desired behavior.

## Proposed behavior

Add a human-invoked `/update` command to the terminal interfaces that checks for a release and offers to install it for future launches. Keep the current process and conversation running throughout. Do not quit, exec, relaunch, automatically resume, replay launch arguments, or change permission settings.

After installation, clearly report: **Update installed for the next launch. This session is still running its original version; you can keep working.** Integrate with #769 to distinguish running, installed, and latest versions. `/new` and `/resume` must not claim to activate the update. A later normal launch uses the updated executable; resuming a saved conversation does not select the old executable.

## Safety and implementation requirements

- Require explicit human authorization before installation. Model output, tool calls, saved conversations, and permissive execution modes must not count as that authorization.
- Use a nonfatal, noninteractive update service, not a direct call to the existing CLI entry point with process-fatal errors and inherited terminal input. Keep progress and failures inside the UI.
- Resolve the actual supported installation target; never execute a project-local or PATH-shadowed updater. Explain unsupported or package-managed installations rather than guessing or escalating privileges.
- Pin the selected release, verify the artifact against the release's trusted authenticity/integrity mechanism, and install atomically. Fail closed if verification is unavailable or fails; do not introduce arbitrary remote-script execution as the in-session update mechanism.
- Serialize concurrent installs. Failure or cancellation before commit must leave the existing installation usable and the live session untouched. Do not downgrade implicitly or reinstall merely because the running process is older than the already-installed version.

## Acceptance checks

Cover successful installation, already-installed/current versions, offline and malformed responses, failed verification, permission errors, cancellation, and simultaneous requests. Verify that the process, conversation, active work, and permissions remain unchanged; a subsequent normal launch reports the installed version. Use isolated installation fixtures, never the developer's live executable.

This removes the extra terminal/install workflow, not the executable lifecycle boundary: activating new code in the current process without a restart is out of scope.

justrach:
Closes #773.

`/update` on the line REPL and TUI checks GitHub for a release and, after an explicit human confirm, installs it for future launches. This process, conversation, and permissions stay on the original binary. `/new` and `/resume` do not activate it.

After a successful install the session reports:

> Update installed for the next launch. This session is still running its original version; you can keep working.

## Safety

- Authorization is a TTY picker or `/update install` after a human `/update` check. Model output, tool calls, saved conversations, and yolo are not consent.
- In-process, nonfatal service — not `graff update` / `install.sh`, no inherited TTY, no exec or relaunch.
- Resolves the real user install (`$HOME/bin/graff` or the running exe). Package-managed and zig-out paths are explained, never overwritten, never PATH-shadowed.
- Pins the selected tag, verifies the tarball against that release's `SHA256SUMS`, installs atomically. Missing or failed verification refuses the install.
- Concurrent installs serialize. Cancel or failure before commit leaves the existing file and the live session untouched. Already-installed current-or-newer disk versions are not reinstalled or downgraded.

Running / installed / latest comparison is shared with `graff update --check` in `version_status.zig`. ADR 0083 records the next-launch boundary (same spirit as ADR 0079).

Acceptance tests use isolated fixtures only and cover success, already-installed, offline/malformed, failed verification, permission errors, cancellation, simultaneous requests, and unchanged process/conversation/permissions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ current is part of cutting a release.
 
 ## Unreleased
 
+- `#773`: `/update` checks a GitHub release and, after an explicit human
+  confirm, installs it for the next launch. The live process, conversation,
+  and permissions stay on the original version; `/new` and `/resume` do not
+  activate it. The in-session service pins the tag, verifies SHA256SUMS, and
+  writes atomically — it does not run `install.sh` or restart. `graff update
+  --check` shares the same running/latest comparison.
 - Prompt-cache affinity is the git root, or one scratch seed when there is
   no `.git`. Eval sandboxes and worktrees can reuse a warm system+tools
   prefix instead of paying it on every leaf cwd. An offline test fails if
--- a/TUI/bgop.zig
+++ b/TUI/bgop.zig
@@ -72,11 +72,13 @@ pub fn finish(self: *Model) void {
             .compact => "compaction failed to start",
             .bash => "command failed to start",
             .files => "file list failed to start",
+            .update => "update failed to start",
         }) catch {};
     } else switch (op.kind) {
         .compact => applyCompact(self, op),
         .bash => applyBash(self, op),
         .files => applyFiles(self, op),
+        .update => applyUpdate(self, op),
     }
     release(self, op);
     self.scroll = 0;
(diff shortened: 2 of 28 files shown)


justrach:

PR #782, opened 2026-09-07
Fix Computer Use form elicitation for read-only get_app_state

### LINKED ISSUE #768: Computer Use bridge fails when MCP client lacks form elicitation
## Summary

On Codegraff v0.0.291, attempting a read-only Computer Use app-state inspection through the documented `@oai/sky` node_repl bridge fails before returning application state.

## Reproduction

1. Load the Computer Use skill.
2. Initialize `@oai/sky` in `mcp__node_repl__js`.
3. Call `sky.get_app_state({ app: "Codegraff", disableDiff: true })`.

## Actual result

```text
nodeRepl.createElicitation is unavailable because the MCP client does not support form elicitation
```

## Expected result

A read-only app-state request should either return the accessibility state, use a supported confirmation path, or provide an actionable fallback when form elicitation is unavailable.

## Impact

Agents cannot visually verify a launched desktop application through the documented Computer Use bridge in clients without form-elicitation support.

justrach:
## Summary

`@oai/sky` `get_app_state` through the signed Codex `node_repl` bridge failed before returning accessibility state:

```text
nodeRepl.createElicitation is unavailable because the MCP client does not support form elicitation
```

Graff advertised empty MCP client capabilities, so node_repl never sent `elicitation/create` and crashed the JS call instead.

## Fix

- Advertise MCP form elicitation on the legacy `initialize` handshake and on modern `_meta.clientCapabilities`.
- During a stdio `tools/call`, answer mid-call `elicitation/create`:
  - accept read-only `sky.get_app_state` (including `disableDiff: true`) with schema defaults;
  - decline URL mode and mutating sky input.
- If `createElicitation is unavailable` still comes back, rewrite it to an actionable fallback that names `disableDiff: true` reconnect (`/mcp trust`) and the desktop `computer` tool (`snapshot` / `apps`). Do not use the plugin's raw Computer Use client or synthesize OS events (ADR 0036, ADR 0071).
- Computer Use skill adapter documents the inspect path and the desktop fallback.

Closes #768.

## Tests

- New reachable tests in `src/mcp_elicitation.zig` (capability advertisement, inspect vs mutate, accept/decline, error rewrite, stdio reply).
- Integration: stdio fixture emits `elicitation/create` mid-`tools/call` for `get_app_state`; the client accepts and the call completes.
- Skill adapter mentions `disableDiff: true` and the desktop `computer` fallback.
- `scripts/eval-tier1.sh --only tests`: 2027 tests (baseline 2009). fmt, lines, reach, build, invariants, sdk green.
- tuiguard `test-tui-hover.py` is unrelated and flaky in this environment; it passed on a solo retry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ current is part of cutting a release.
 
 ## Unreleased
 
+- `#768`: the Computer Use `node_repl` bridge no longer dies on
+  `createElicitation is unavailable` during a read-only
+  `sky.get_app_state({ disableDiff: true })`. Graff advertises MCP form
+  elicitation, accepts that inspect mid-call, and otherwise returns an
+  actionable fallback that names the desktop `computer` tool. ADR 0036,
+  ADR 0071.
 - Prompt-cache affinity is the git root, or one scratch seed when there is
   no `.git`. Eval sandboxes and worktrees can reuse a warm system+tools
   prefix instead of paying it on every leaf cwd. An offline test fails if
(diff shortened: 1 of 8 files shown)


justrach:

PR #787, opened 2026-09-08
fix: worktree edits land on the selected root (#747, #748, #750)

### LINKED ISSUE #747: Workspace edits can report success without updating the selected worktree
After switching worktrees, an edit can report backend success while the harness rejects it with: "the edit ... did NOT persist" and "the file on disk is byte-for-byte what it was before."

The persistence check prevents a false success, but the edit does not land. Suspected cause: the edit backend and read-back verification resolve different workspace roots; this still needs confirmation.

Ensure both paths use the selected worktree and add a regression that switches between two worktrees before editing the same relative filename.

### LINKED ISSUE #748: Error-only SSE responses are retried as truncated bodies
An explicit streamed error is reported as an unparseable response after truncated-body retries.

The error-only SSE path fails response reconstruction instead of surfacing the error event. Recognize error events before the truncated-response retry classifier, preserve their message, and avoid retrying deterministic invalid-request errors.

### LINKED ISSUE #750: Resize-anchor PTY guard intermittently fails in the parallel probe suite
The resize-anchor guard failed during the parallel pre-push probe suite, then passed unchanged in isolation. The failing assertion reported excessive marker-row drift during a width change. The cause is not confirmed; investigate timing dependence in the probe before changing anchor behavior or relaxing the assertion.

justrach:
Closes #747. Closes #748. Closes #750.

Targets the current `release/v0.0.292` tip (`5c22fca`).

## What was still missing

**#747** was the only one of these three that was not on 292. After `/workspace use`, `edit_file` / `write_file` / `read_file` could report success while verification looked at a different root (stale Threaded-Io cwd after posix `chdir`, or a display cwd that disagreed with posix).

`sessionAbs` now resolves one absolute path for write and verify: `agent_cwd`, then posix process cwd, then display cwd — never `Io.Dir.cwd()`.

## Already on 292 (closing via this PR)

- **#748**: error-only OpenAI SSE is an error envelope, not a truncated-body retry (`sse-error-not-truncated`).
- **#750**: resize-anchor waits for a stable frame after a width change. The PTY probe passed in this run (`widths/rows` held the marker).

## Tests

- New `src/edit_worktree_tests.zig` (3 cases), including the required invariant `edit-selected-worktree`.
- Wired `agent_request_encrypted_tests.zig` into the test root so the 292 encrypted-reasoning backtest is actually compiled (tier-1 reach was red on tip without this).
- `scripts/eval-tier1.sh` green: 2069 tests (baseline 2063), all 18 TUI probes, including resize-anchor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ current is part of cutting a release.
   width change instead of asserting on the first (possibly stale) repaint.
 - `#748`: an error-only SSE event is reported as that error, not retried
   as a truncated gateway body.
+- `#747`: `edit_file` / `write_file` / `read_file` share one absolute path
+  from the posix session cwd (not a stale Io cwd or a disagreeing display
+  cwd), so a worktree switch cannot report success on the previous tree.
 - `#761`: `read_file` optional fields stay out of `required[]` (path only).
   Null or empty `contains` / null bounds / null `compact` are a whole-file
   read, so strict tool calling can omit unused mutually exclusive fields.
--- a/scripts/eval/tier1-manifest.json
+++ b/scripts/eval/tier1-manifest.json
@@ -72,6 +72,11 @@
       "id": "sse-error-not-truncated",
       "test": "#748: error-only OpenAI SSE is an error envelope, not a missing stream",
       "why": "#748: an explicit streamed error was retried as a truncated gateway body."
+    },
+    {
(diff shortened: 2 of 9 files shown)


justrach:
The remaining selected-root fix is integrated into release/v0.0.293 as 0889701a. The other fixes were already present. Release checks and targeted file/workspace behavior checks passed.

PR #795, opened 2026-09-08
Fix TUI pasted item boundaries and file spans

### LINKED ISSUE #792: TUI: separate pasted items and make pasted file paths atomic spans
## Requested behavior

Make paste-originated composer items keep a safe boundary and preserve their identity as pasted objects.

- After a completed text paste, pasted-text marker, file paste/drop, or image insertion, leave one trailing space unless whitespace is already present. The user can backspace over that separator when they want concatenation.
- Repeated file pastes or drops should remain separate, for example `/path/to/a /path/to/b`, never `/path/to/a/path/to/b`.
- A file path inserted by paste or drag/drop should be a highlighted semantic span, extending the pasted-text behavior from #673: backspace/delete at its edge removes the whole inserted file item, and word navigation does not enter or partially break it.
- A path typed manually must remain ordinary editable text. Paste/drop origin, not matching a path-shaped string, should determine whether it is atomic.
- Removing an atomic file span must also remove any associated attachment/reference state.

## Why

Successive pastes and typing can currently join directly onto the previous pasted content. For file paths, this can turn two valid paths into one invalid path and cause file retrieval to target the wrong location.

Pasted or dropped files are also conceptually single composer objects, like image and pasted-text markers, but character-by-character editing can corrupt the path while it still appears to represent the pasted file.

## Acceptance checks

1. Paste `hello`, then type `world`: the composer contains `hello world` unless the inserted separator was explicitly removed.
2. Paste or drop two files consecutively: two distinct paths remain separated.
3. Backspace/delete at the boundary of a pasted file span removes the whole span; word navigation skips over it.
4. Manually type the same path: it remains editable character by character.
5. A paste that already ends in whitespace does not gain duplicate separator whitespace.

yxlyx:
## What changed

- Track each bracketed paste as an exact composer range and append one removable trailing space unless the pasted content or following text already supplies whitespace.
- Normalize pasted or dropped local file paths into origin-backed semantic spans. Backspace, Delete, arrow/word navigation, range deletion, and undo now treat those spans atomically, while manually typed identical paths remain ordinary text.
- Preserve live span identity while temporarily navigating prompt history, and style only origin-backed spans in the composer.
- Keep image attachments on the existing chip path while consuming only the image paste range instead of clearing unrelated draft text.
- Add focused #792 coverage and update existing paste-trajectory expectations for the new separator contract.

## Why

Successive pastes could concatenate text and file paths into invalid input, and pasted file paths could be damaged one character at a time despite representing a single inserted item. Capturing the paste range at insertion time preserves origin without guessing from path-shaped text later.

The separator remains ordinary text so users can intentionally remove it. File-span classification requires a real local filesystem entry; image paths retain their existing attachment behavior. Prompt history stores submitted lines as plain text, while only the live unsent draft preserves semantic span state.

## Verification

- `zig build tui-test '-Dtest-filter=#792'`
- `zig build tui-test` — 569/569 passed
- Tier 1 formatting, line ceiling, reachability, build, TUI, PTY guard, invariants, and SDK checks
- `zig build test --summary all` — 2071 passed, 1 skipped, 2072 total
- `python3 scripts/eval/tier1_test_binary.py count` — 2072

Fixes #792

### DIFF
--- a/TUI/app.zig
+++ b/TUI/app.zig
@@ -84,7 +84,7 @@ pub const Model = struct {
     prompt_hist: std.array_list.Managed([]const u8),
     /// Image paths index-aligned with `prompt_hist` (#577). Empty slice = text-only.
     prompt_hist_images: std.array_list.Managed([]const []const u8),
-    draft_text: ?[]u8 = null,
+    draft_input: ?input_mod.State = null,
     draft_images: ?[]const []const u8 = null,
     steer_queue: std.array_list.Managed([]const u8),
     images: std.array_list.Managed([]const u8),
--- a/TUI/chrome.zig
+++ b/TUI/chrome.zig
@@ -49,7 +49,7 @@ pub fn promptBox(self: *const Model, a: std.mem.Allocator, width: usize) ![]cons
         try body.appendSlice(try imageChips(self, a, th.accent, th.text));
         try body.appendSlice("  ");
     }
-    try body.appendSlice(try self.input.view(a));
+    try body.appendSlice(try self.input.viewStyled(a, th.accent, th.text));
     const wrapped = try theme_mod.wrapPreferWords(a, body.items, inner);
     var lines = std.array_list.Managed([]const u8).init(a);
     var it = std.mem.splitScalar(u8, wrapped, '\n');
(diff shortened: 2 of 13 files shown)


justrach (changes requested):
normalizePath percent-decodes ordinary filesystem paths as well as file URLs. When both literal-percent and decoded filenames exist, this can select the wrong file. Decode percent escapes only for file URLs and add a regression covering that collision before merging.

justrach:

justrach:
Accepted: merged into release/v0.0.293 as 84f26cda (branch tip c010765f), verified with the full unit + TUI suites and tier-1 checks.

PR #796, opened 2026-09-08
TUI paste/URL spans, constraint scope, ACP Stop, failed-subagent evidence

### LINKED ISSUE #788: TUI composer splits wrapped URLs into partial modifier-click targets
## Symptom

A supported safe link typed or pasted into the TUI composer becomes only partially clickable when it wraps onto a second visual row. Modifier-hover underlines only the fragment before the wrap, and modifier-click opens that truncated prefix instead of the original link.

This affects explicit URLs such as `https://example.com/...` and should also be handled for other safe link forms the terminal recognizes, such as `www.example.com/...`.

## Root cause

`promptBox` passes raw composer text to `wrapPreferWords`, which inserts physical newline-separated rows. The composer does not emit OSC 8 hyperlink metadata for recognized link tokens, so terminal link detection treats each visual row independently.

## Expected behavior

Any supported safe link that wraps visually should remain one hyperlink whose target is the original unbroken link across every continuation row.

## Fix

Detect supported safe link forms before wrapping—at minimum `http://`, `https://`, and `www.`-style addresses—and emit an OSC 8 target for the complete token, ensuring wrapped continuation rows retain the same target. Keep detection allowlisted so arbitrary or unsafe URI schemes are not made clickable. Add headless TUI coverage for both an explicit-scheme URL and a `www.`-style address longer than the composer row width.

### LINKED ISSUE #789: Constraint capture should not turn temporary steering into project-wide policy
## Problem

Constraint capture currently conflates three different kinds of user intent:

- an immediate correction ("not this right now")
- a preference limited to the current task or session
- a durable rule for every future task in the project

The capture instruction tells the agent to record a standing constraint whenever the user rejects, forbids, or vetoes something. The resulting entry is unscoped, append-only, and injected as a hard constraint into future sessions and workers until the user explicitly retires it.

As a result, contextual steering can be rewritten into a project-wide prohibition the user never intended. Over time, users accumulate hard constraints with unclear origins and must manually discover and remove them.

## Proposed direction

1. Give recorded guidance an explicit scope such as `turn`, `task`, `session`, or `project`.
2. Treat an implicit rejection as local to the current task/session by default. Only create a durable project constraint when the user clearly states standing intent (for example, "never do this in this project") or confirms that scope.
3. Preserve limiting language such as "here", "for this task", and "right now" instead of generalizing it into an unconditional imperative.
4. Make automatic capture visible and immediately reversible, with the recorded text and scope shown to the user.
5. Make the constraint review UI show origin, scope, and creation context, and support straightforward cleanup of incorrectly captured entries.
6. Provide a migration/review path for existing unscoped auto-captured constraints rather than silently treating all of them as permanent project policy.

Improving removal alone is not sufficient: the primary failure occurs when temporary steering is classified and persisted as durable policy.

## Acceptance criteria

- "I don't want this right now" does not create a project-wide standing constraint.
- "For this task, do not do X" expires with the task/session scope.
- "Never do X in this project" can create a durable project constraint.
- Ambiguous intent either remains local or asks for confirmation before durable capture.
- Users can see why, when, and at what scope each active constraint was recorded, then undo or retire it without knowing an internal ID.
- Behavior tests cover temporary preferences, task-scoped instructions, explicit standing rules, and language whose scope must not be widened.

### LINKED ISSUE #791: ACP Stop waits for the active turn to finish
Stop does not interrupt an active ACP turn because synchronous dispatch blocks reading cancellation notifications. The GUI also hides cancellation failures and can immediately start queued work.

Fix: read cancellation independently of turn dispatch, preserve turn-scoped cancellation state, surface failures, and pause queued work. Add a distinct Steer action that stops and continues the same session.

### LINKED ISSUE #792: TUI: separate pasted items and make pasted file paths atomic spans

### LINKED ISSUE #793: Failed sub-agents omit useful partial evidence
A sub-agent can perform useful work and then fail, but its parent receives only the failure cause. The error-return path omits retained findings and tool results; workflow excerpts also discard that evidence. Preserve bounded, explicitly incomplete evidence while keeping error status intact, so the parent can summarize findings and unfinished work honestly.

justrach:
Fixes five open issues on `main` (v0.0.293). None of these were already solved.

### #788 — wrapped composer URLs stay one click target
`promptBox` now linkifies allowlisted `http://`, `https://`, and `www.` tokens as one OSC 8 span before `wrapPreferWords`. Continuation rows keep the original href. `[Image]` / `[Pasted text` chips and other schemes (`javascript:`) are left alone.

### #792 — paste separators and atomic file spans
A completed text paste, file drop, or file-picker insert leaves one trailing space unless whitespace is already present. Consecutive file inserts stay `/a /b`, never `/a/b`. Paste/drop origin makes a path an atomic span (backspace/word-nav); a typed lookalike stays ordinary text.

### #789 — constraint capture is scoped
`note_constraint` records `turn` / `task` / `session` / `project`. Implicit rejections default to session. “Right now” is turn-local and is not written as project policy. “Never … in this project” (or `scope=project`) is durable. Legacy unscoped ledger rows still inject and show as review-needed in `/never`. ADR 0088.

### #791 — ACP Stop does not wait for the turn
`session/cancel` is applied from a stdin reader while `session/prompt` is still in dispatch. Cancel state is per-dispatch so a prior stop cannot cancel the next prompt. The GUI surfaces cancel failures, holds the follow-up queue on Stop, and adds Steer (cancel, then continue the same session). ADR 0089.

### #793 — failed sub-agents keep partial evidence
The error-return path appends bounded, labelled findings and tool names (`[partial evidence, incomplete]`). Workflow excerpts include that section so the parent can summarize unfinished work without losing error status.

Closes #788. Closes #789. Closes #791. Closes #792. Closes #793.


justrach:
Closing in favor of the single-issue PRs covering the same fixes: #801 (wrapped composer links, #788), #795 (paste separators and atomic file spans, #792), and #799 (scoped constraint capture, #789). The remaining unique parts (ACP Stop #791, failed sub-agent evidence #793) are tracked on the v0.0.293 release branch.

PR #799, opened 2026-09-08
Require explicit project scope for constraints

### LINKED ISSUE #789: Constraint capture should not turn temporary steering into project-wide policy

yxlyx:
## What changed

- Require explicit `scope: project` and a complete verbatim sentence or line from the current user message before durable constraint capture.
- Write scoped records, mark legacy unscoped records for review, promote exact legacy or learned matches append-only, and reject normalized-text collisions.
- Expose scoped review and removal through `/never` in the fullscreen TUI and add regression coverage.

## Why

Temporary steering could be widened into durable project policy, while an existing legacy or learned match could be reported as confirmed without actually changing its scope. The new path fails closed instead of guessing intent from language or session lifecycle, while keeping existing records active and visibly reviewable for compatibility.

Verified with unit, TUI, PTY guard, scripted behavior, reachability, formatting, build, and SDK consistency checks.

Fixes #789


yxlyx:
@justrach This changes the constraint-persistence contract rather than only applying a narrow bug fix, particularly around explicit project scope and compatibility treatment for legacy unscoped records. Could you please review the product and compatibility decision before merge?

justrach:

justrach:
Accepted: merged into release/v0.0.293 as 113495d2 (branch tip c010765f), verified with the full unit + TUI suites and tier-1 checks.

PR #801, opened 2026-09-08
fix(tui): preserve wrapped composer link targets

### LINKED ISSUE #788: TUI composer splits wrapped URLs into partial modifier-click targets

yxlyx:
## What changed

- Detect allowlisted `http://`, `https://`, and `www.` links in the TUI composer and emit OSC 8 targets that remain balanced across wrapped rows.
- Retain only the visible composer tail before repeating link metadata, keeping large pasted links bounded.
- Preserve UTF-8 cursor, deletion, and undo boundaries while editing linked drafts.
- Add reachable headless coverage for wrapping, resizing, truncation, cursor placement, malformed targets, and unsafe schemes.

## Why

The composer inserted physical newlines into long links without hyperlink metadata, so terminal modifier-click opened only the fragment before the wrap. The fix is composer-specific to avoid changing transcript wrapping, and it uses an explicit web-scheme allowlist so arbitrary URI schemes are not made clickable. Each retained row closes and reopens the same logical target before padding and borders, which prevents hyperlink state from leaking during independent row paints.

## Testing

- `zig build tui-test --summary all` — 566/566 passed
- `scripts/eval-tier1.sh` — format, line limit, spec, reachability, build, TUI, all PTY probes, invariants, and SDK checks passed; the general suite-count check remains red at 2071 vs 2072 and reproduces unchanged on `origin/main`

Fixes #788


justrach:

justrach:
Accepted: merged into release/v0.0.293 as e15eef67 (branch tip c010765f), verified with the full unit + TUI suites and tier-1 checks.

PR #821, opened 2026-09-10
fix(repl): keep image chips atomic and renumbered

### LINKED ISSUE #820: Line REPL image chips can split or repeat stale ordinals
## Problem

In the terminal composer, modified word deletion can split an `[Image #N]` chip into literal text. Repeated staging of an identical image can also leave stale `[Image #2]` chips, so multiple chips display the same ordinal.

## Root cause

Word-wise cursor and deletion paths bypass image-chip boundaries. Separately, payload deduplication can remove a queue slot while the zero-remapped chip remains in the editable buffer.

## Fix

Route word motion and deletion through chip-aware boundaries. Remove zero-remapped chips only when they were backed by composer attachments, then reuse live ordinals. Add regressions for both paths.

### LINKED ISSUE #834: SDK packed-install smoke can resolve a released binary instead of its fixture

### LINKED ISSUE #835: Linux subagent cancellation test iterates a directory without iteration enabled

yxlyx:
Fixes #820
Fixes #834
Fixes #835

## What changed

- Route modified word motion and deletion through image-chip boundaries.
- Remove stale composer-backed chips when payload deduplication drops their queue slot.
- Generate packed-SDK native fixtures at the optional-dependency versions they represent and require the consumer install to stay offline.
- Open the subagent cancellation test directory with iteration enabled.
- Teach WebSocket recovery assertions to accept only the replacement socket's observed prewarm anchor.
- Add regressions for atomic word deletion and reused live ordinals.

## Why

Image chips are attachment controls, not ordinary words. Generic word boundaries could split a chip into literal text. Separately, payload deduplication could remove an attachment slot without removing its chip, allowing repeated `[Image #2]` labels to accumulate.

The Linux CI fixture also reached directory pruning through a handle that was not opened for iteration, causing `BADF`. The SDK package smoke built native fixtures at the SDK version rather than the versions declared by its optional dependencies, which allowed package resolution to select a released binary instead of the isolated fake executable.

WebSocket recovery can legitimately prewarm a replacement connection before replaying full history. The previous assertion rejected every parent anchor, misclassifying that fresh connection-local prewarm as a stale inference chain.

## Constraints and trade-offs

- Zero-remapped chips are removed only when they were backed by composer attachments, so literal typed `[Image #N]` text remains unchanged.
- Existing image-payload deduplication behavior is preserved.
- SDK and native package releases may differ; the test follows the declared dependency pins instead of changing published versions.
- Offline fixture installation makes accidental registry fallback fail directly rather than requesting credentials.
- WebSocket assertions still reject inference parents and anchors from other connections, and still require recovery to use a new socket with full input history.

### DIFF
--- a/scripts/codex_ws_error_test.py
+++ b/scripts/codex_ws_error_test.py
@@ -188,11 +188,15 @@ def run_chain_reanchor_scenario(
             first.connection_id != rejected.connection_id
             or rejected.connection_id == rebuilt.connection_id
             or rejected.body.get("previous_response_id") != "resp_chain_1"
-            or "previous_response_id" in rebuilt.body
+            # A fresh socket may chain only from its observed prewarm,
+            # never from the rejected inference response or another socket.
+            or not mock.has_fresh_parent(rebuilt)
         ):
             raise AssertionError(
                 "chain-error: rejected delta was not rebuilt as full input on a fresh WS: "
-                f"{requests!r}"
+                f"connections={(first.connection_id, rejected.connection_id, rebuilt.connection_id)!r}, "
+                f"rejected_parent={rejected.body.get('previous_response_id')!r}, "
+                f"rebuilt_parent={rebuilt.body.get('previous_response_id')!r}"
             )
         rebuilt_types = [
             item.get("type")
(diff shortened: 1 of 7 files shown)


justrach:

Issue #817, opened 2026-09-10
Upgraded Graff misses pre-registry orphan listeners and allows split IPv4/IPv6 port ownership


yxlyx:
## What happened

On `graff 0.0.295`, startup and `graff servers` failed to surface a localhost listener that had originally been launched from a Graff tool environment on August 28, 2026, before the ownership registry landed in PR #732.

When discovered on September 10, 2026, the listener:

- had been reparented to PID 1 because its launcher/session was gone;
- was the leader of its own process group;
- retained a Graff-specific environment marker;
- had remained bound to `127.0.0.1:3000` for more than 12 days; and
- was absent from `graff servers` because it had no registry record.

A new Graff-owned background server was then able to bind `*:3000` through IPv6. The result was silent split ownership: `127.0.0.1:3000` and `[::1]:3000` served different applications with no warning. The stale listener was terminated only after explicit user authorization.

## Why this is still a gap

PR #732 addresses jobs registered after the lifecycle feature shipped, but machines upgraded from older versions can retain pre-registry orphan listeners indefinitely. Issue #321 also documented `ORPHANED_OWNED_JOB` and `LISTENER_AFTER_CLEANUP` as parked work, but it is now closed.

This report does not claim that a newly registered `0.0.295` job leaked. It covers reconciliation of older Graff-originated orphans and detection of same-port collisions across address families.

## Expected behavior

At startup and in `graff servers` or `/doctor`:

1. Warn about orphan listeners with strong Graff provenance even when no current registry JSON exists. Evidence could combine a Graff-specific environment marker, PPID 1, process start identity, and a listening socket.
2. Never auto-kill from a weak heuristic. Offer a validated stop action that rechecks PID, start identity, executable, and process group.
3. Before starting a managed server, check the requested port across IPv4 and IPv6. Warn or fail rather than silently allowing different applications to own the same numeric port on different loopback families.
4. Consider a one-time post-upgrade reconciliation path for processes created before PR #732.

## Reproduction outline

1. Using a build from before PR #732, launch a long-running localhost server from a Graff bash tool and let its owning session exit without removing the descendant.
2. Upgrade to `0.0.295`.
3. Run `graff servers`; the orphan is absent because it has no ownership record.
4. Start another managed server on the same numeric port, bound through the other address family.
5. Observe that IPv4 and IPv6 loopback URLs resolve to different applications without a Graff warning.

Related: #199, #198, #321.

yxlyx:
[...]

PR #822, opened 2026-09-10
fix: safely surface orphan listeners and guard server cleanup

yxlyx:
## What changed

- Add bounded, read-only orphan-listener discovery to `graff servers`, interactive startup, and `/doctor`. Suspected legacy listeners remain separate from verified ownership records.
- Reject missing process identities and unsafe groups when stopping recorded jobs. Revalidate before escalation and distinguish failed probes or signals from confirmed termination, including zombie-only groups.
- Preflight recognizable server commands against TCP listeners across IPv4 and IPv6 without stopping the incumbent.
- Preserve jobs with listening sockets or unknown socket state during automatic idle and session cleanup.

## Why

Pre-registry listeners were invisible to the ownership-record scan, and separate address families could silently serve different applications on the same numeric port. Silence also does not establish that a browser tab is closed. These changes improve visibility and prevent unsafe automatic cleanup without turning inherited environment hints into stop authority.

## Constraints and trade-offs

- Suspected legacy listeners are report-only, not automatically adopted or stopped. Environment markers and ancestry cannot establish ownership.
- Browser visibility remains incomplete. Listening or unverifiable jobs are retained conservatively; automatic cleanup after the final browser tab closes is not implemented.
- Port preflight recognizes conservative command forms with explicit numeric ports. Unsupported shell syntax is not covered; failed probes warn explicitly. Preflight is not a socket reservation and cannot eliminate the bind race.
- Recorded stop checks use start identity and process-group leadership. Existing records lack executable fingerprints, and userspace revalidation is not an atomic kernel ownership guarantee.

## Verification

- Full `scripts/eval-tier1.sh` passed: formatting, source limits, test reachability, build, unit tests, TUI checks, invariants, and SDK drift.
- Added deterministic discovery, browser-protection, and stop-race regressions plus a real loopback port check.
- Fixed the retained-job regression uncovered by the combined suite.

Refs #817.

## CI follow-up

The Linux cancellation fixture now opens its directory for iteration, matching the production recorder. The packed SDK smoke creates native mocks at the declared dependency versions and installs offline, preventing accidental resolution of a real credential-requiring binary. WebSocket recovery assertions distinguish a new connection's observed prewarm anchor from a stale inference chain while retaining fresh-socket and full-history checks.

All six GitHub checks passed after these repairs: SDK, Linux Zig, and Windows, on both push and pull-request workflows.

### DIFF
--- a/docs/adr/0096-orphan-listeners-are-not-stop-authority.md
+++ b/docs/adr/0096-orphan-listeners-are-not-stop-authority.md
@@ -1,0 +1,44 @@
+# 0096. Orphan listener evidence is not stop authority
+
+Status: accepted 2026-09-10
+
+## Context
+
+Ownership records introduced by ADR 0062 cannot describe servers that were
+started before the registry existed or whose record write failed. Inherited
+environment variables and an orphaned process group can help discover those
+listeners, but cannot prove that Graff may safely terminate them. Separate
+IPv4 and IPv6 listeners can also use the same numeric port.
+
+## Decision
+
+Discover probable legacy listeners as read-only candidates. Never adopt them
+into the ownership registry or signal them based on an environment marker.
+Permission failures and incomplete probes remain explicit uncertainty.
+
+For recorded jobs, stopping requires a nonzero matching process start identity
+and a verified dedicated process group. Revalidate before escalation; failed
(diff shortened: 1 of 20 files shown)


justrach:

PR #831, opened 2026-09-10
fix: keep chat conversations vertically scrollable

### LINKED ISSUE #827: GUI chat should scroll only vertically, not sideways
## Problem
GUI conversations can move horizontally, shifting message text out of view. The conversation should scroll only up and down, including in narrow chat panes.

## Suspected cause
The transcript container enables vertical scrolling without explicitly containing horizontal overflow. Wide message content can overflow the conversation viewport.

## Requested fix
Constrain the conversation to its available width and prevent horizontal scrolling of the chat viewport. Wrap prose, links, and inline code as needed; contain genuinely wide content within its own block rather than widening the conversation. Do not merely clip unreadable content.

## Acceptance
- Horizontal and diagonal trackpad gestures do not shift the conversation sideways.
- Vertical scrolling continues to work normally.
- Long links, inline code, code blocks, and tables do not widen the chat viewport or push message text out of view.
- Narrow panes and resizing preserve readable content and a fixed horizontal position.

### LINKED ISSUE #834: SDK packed-install smoke can resolve a released binary instead of its fixture

### LINKED ISSUE #835: Linux subagent cancellation test iterates a directory without iteration enabled

### LINKED ISSUE #837: Desktop composer skill-selection regression fails under pointer input
The desktop interaction suite repeatedly fails with:

`A portaled skill selection must reach the composer`

Selecting a skill through pointer input leaves the composer empty. The fixture scrolls and immediately measures the menu option before sending pointer input, while menu placement and active-option scrolling can still update. A missed target is a possibility, not a confirmed cause.

Determine whether input misses the option or the selected draft is subsequently cleared, then fix the responsible path while preserving the end-to-end assertion.

### LINKED ISSUE #846: Fresh-connection transport tests reject valid warm-up anchors

yxlyx:
## What changed
- Constrain the chat viewport and wrap long prose, links, and inline code.
- Preserve local horizontal scrolling for wide code blocks and tables.
- Add gesture, resizing, and narrow-pane regressions to desktop CI.
- Repair separate package, cancellation, pointer-targeting, and transport-test blockers in separate commits.

## Why
Unbroken content could widen the transcript and let conversations shift sideways. Wrapping text and containing wide blocks keeps content readable; hiding overflow alone would leave text clipped. The change preserves normal vertical scrolling without intercepting input events.

The package smoke matches fixture versions to dependency pins and installs offline. The cancellation fixture enables directory iteration. Pointer tests wait for stable hit-tested targets. Transport checks distinguish a fresh connection’s observed warm-up anchor from a stale inference chain, retaining connection-replacement and full-history assertions. These test-only repairs unblock validation without changing published versions or runtime behavior.

Closes #827.
Closes #834.
Closes #835.
Closes #837.
Closes #846.

### DIFF
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -22,6 +22,7 @@ jobs:
       - run: bun install --frozen-lockfile
       - run: bun run test:desktop
       - run: bun run build
+      - run: bun run test:overflow
       - run: bun run test:projects
       - run: bun run test:stress
       - run: bun run test:agents
--- a/apps/native/app/visual-tests/overflow/overflow-fixture.tsx
+++ b/apps/native/app/visual-tests/overflow/overflow-fixture.tsx
@@ -1,0 +1,126 @@
+"use client";
+
+import { useCallback, useMemo, useState } from "react";
+import ChatTranscript from "@/components/site/ChatTranscript";
+import { emptyTurn, type AssistantTurn } from "@/lib/acp";
+import type { Msg } from "@/components/site/harness-types";
+
+const noop = () => {};
+const widths = [260, 360, 720] as const;
+const cases = ["all", "url", "prose", "inline-path", "nested", "table", "fence", "stream", "error", "recap", "reasoning", "user"] as const;
+type Case = typeof cases[number];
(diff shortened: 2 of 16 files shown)


justrach:

PR #854, opened 2026-09-10
fix: TUI follow-up spinner and isolated probe clipboard (#845, #836)

justrach:
Fixes #845 and #836.

- Queuing a TUI follow-up no longer freezes the thinking indicator: `spinnerTask` keeps frames while the steer row is visible, hopping one line with DECSC/DECRC so the input is not wiped.
- Tuiguard probes get private copy/paste via `GRAFF_CLIPBOARD_COPY` / `PASTE` so parallel clipboard tests do not collide.

Also unblocks Linux `zig build test`: the cancel-file unit test now iterates its tmpDir (same flag as the retention test).

Branch is off latest `main`. Tier 1 green, including 19 tuiguard PTY probes.

<div><a href="https://cursor.com/agents/bc-01a02d64-4949-703b-8c44-a9bb932f4ffc?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a02d64-4949-703b-8c44-a9bb932f4ffc&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #833, opened 2026-09-10
Hide task disclosures when no details are available

yxlyx:
## What changed
- Render detail-free task rows as static content, without disclosure controls or empty detail spacing; preserve expansion when details exist.
- Add rendered and interactive task-row regressions and run the browser suite in desktop CI.
- Repair the offline SDK package fixture and Linux cancellation fixture.
- Stabilize composer pointer tests by settling hover/layout and hit-testing before pressing.
- Distinguish a fresh socket's observed prewarm anchor from a stale response chain in transport tests.

## Why
Task labels and statuses do not imply details exist. Disclosure must depend on actual content, not progress or completion.

CI verification also exposed fixture defects: mismatched mock package versions could select a released executable, directory pruning required an iterable handle, pointer coordinates could become stale, and valid fresh-connection prewarming was misclassified as stale history. The fixes preserve the selection, socket-replacement, full-history, and stale-chain assertions rather than weakening them.

## Validation
- Desktop unit tests, production build, task/project/stress/agent browser suites, and updater bundle check passed.
- SDK tests, typecheck, build, and offline package smoke passed.
- Zig unit suite and transport scenarios passed.
- Full offline tier-1 pre-push checks passed.
- All seven GitHub checks passed.

Fixes #826. Fixes #837.
Related fixture repairs: #834 and #835.

PR #856, opened 2026-09-10
fix: bash timeout bound, MCP teardown hang, and transport fixture anchors

### LINKED ISSUE #834: SDK packed-install smoke can resolve a released binary instead of its fixture

### LINKED ISSUE #835: Linux subagent cancellation test iterates a directory without iteration enabled

### LINKED ISSUE #846: Fresh-connection transport tests reject valid warm-up anchors

### LINKED ISSUE #848: Transport replay fixtures reject valid fresh prewarm anchors
## Problem
Transport recovery and compaction fixtures reject valid full-history replays on fresh WebSocket connections.

## Cause
The fixtures treat every previous-response anchor as stale, including the empty-input prewarm anchor issued on the new connection.

## Fix
Record prewarm anchors per connection and accept only that connection's anchor. Keep assertions for fresh connections, full replay history, and rejection of stale anchors.

### LINKED ISSUE #849: MCP connection-reuse test can hang during teardown
The MCP connection-reuse test can block indefinitely during cleanup.

Zig executes deferred actions in reverse order. Cleanup opens the wake connection before setting the stop flag, allowing the server to return to accept before the final future await. No further connection releases that wait.

Fix: set the stop flag before waking the listener, then await the server. Add a teardown regression and a bounded test-run deadline so a failed cleanup cannot stall verification indefinitely.

### LINKED ISSUE #850: Explicit bash timeout bypasses foreground auto-background safety
A root `bash` call with a large explicit `timeout` can remain in `· bash still running ·` instead of returning control at the normal auto-background boundary. This recreates the failure mode fixed by #620 when an agent treats `timeout` as a command deadline rather than a foreground wait.

Root cause: `src/exec_bash.zig` `rootWaitMs` replaces the default promotion delay with any positive `timeout`, capped only by `job_wait.wait_cap_ms`. `waitForeground` therefore cannot return `.running` and hand back a job id until that full value expires. The tool schema's “120s (or timeout ms)” wording makes this misuse easy.

Fix: keep root foreground promotion bounded even when `timeout` is supplied. A shorter explicit value may promote earlier, but a larger value should not extend foreground blocking beyond `root_wait_ms`; the process can continue in the job registry. If root commands need a separate lifetime deadline, expose it independently instead of overloading the foreground wait.

Add a regression where a large explicit timeout still promotes at the bounded foreground threshold while leaving the child alive and accessible through `bash_output`/`bash_kill`.

justrach:
Draft off latest `origin/main` (`e5aefb2`). No PR template in the repo.

## Fixes

- **#850** Root `bash` `timeout` may only *shorten* the foreground wait. A large explicit value can no longer stretch `rootWaitMs` toward the 10h job-wait cap. Schema / lean `-p` copy and ADR 0026 match. A promoted child stays on `bash_output` / `bash_kill`.
- **#849** MCP reuse teardown is stop → wake → await, with a 2s extra-wake watchdog. Reverse `defer` order used to re-enter `accept` and hang the suite.
- **#848 / #846** Codex mock records a per-connection prewarm/warmup id. Recovery and compaction fixtures accept no parent *or* that socket's anchor only. Stale-parent, connection-replacement, and full-history checks stay.
- **#835** Cancel-file and coalescing recorder tests open the tmp dir with `.iterate = true` so Linux `prune` is not `BADF`.
- **#834** Packed-install fixtures pass `--version` from the SDK `optionalDependencies` pin and `npm install --offline`, so a miss fails instead of resolving a released binary.

Closes #850. Closes #849. Closes #848. Closes #846. Closes #835. Closes #834.

## Verification

- `scripts/eval-tier1.sh` green (suite 2134; floor 2105 is a stale-ratchet warning only).
- New Zig tests are imported (`test_hooks.zig` → `transport_fixture_tests.zig`; others live in already-imported modules).
- `python3 scripts/codex_ws_anchor_test.py` → `ok`.
- `schema.zig` stays 599 LOC. SDK generate is in sync (description-only bash timeout wording).

## Risks

- Codex PTY transport smokes (`scripts/codex_ws_test.py` / `codex_ws_error_test.py`) are not in tier 1; they need a built `graff` and a PTY.
- Packed-install npm smoke is not in tier 1; `--offline` will fail loud if a fixture version still misses.
- Empty-input + no `previous_response_id` is treated as a prewarm. A real empty first turn on that shape would be answered as setup, not counted as a `ws_turn`.
- #835 also appears in open PR #854 (same `iterate` flag). #834/#848 are mentioned as incidental repairs on unmerged GUI PRs #833/#838; those are not this branch.


justrach:
Superseded by #858 (landed on `release/v0.0.296`) and #859 (main breakage / leftover CI fixes). Closing this overlapping draft from the same issue-wave.

Issue #847, opened 2026-09-10
Non-draft PR publication skips failed-head and claim verification


yxlyx:
## Problem

A session can publish a non-draft pull request even when the exact head commit already has failed branch CI and the PR description overstates what its tests prove. The current publication safeguards sanitize outbound text, but they do not establish that the proposed change is actually ready for review.

This can produce a public PR that is red on arrival, omits known failures and verification evidence, and claims a behavior guarantee that still fails just outside the tested input.

## Reproduction

1. Implement a behavior fix and add a narrow helper-level regression.
2. Run local checks successfully and push the branch.
3. Let push-triggered CI fail for the head commit.
4. Poll only for whether a PR exists; do not inspect the head commit run.
5. Create a non-draft PR whose body says regressions were added, but does not list the commands run, the remote failures, or the untested edge cases.
6. PR-triggered CI runs the same failing commit again.
7. A later direct audit finds that the claimed guarantee is incomplete despite the green local suite.

A representative failure is an atomic-token fix whose helper handles one separator but falls back to generic word parsing with multiple separators. The test asserts one boundary value rather than exercising the actual deletion path, so both the test and PR description look stronger than the behavior they cover.

## Why the later CLI errors are not the cause

`gh pr create` succeeds. A later `gh pr checks --watch` exits non-zero because checks failed, which is expected behavior and arrives too late to protect publication. The defect is the missing readiness decision before the non-draft PR is created.

A failing head run may be an existing main-branch failure rather than a regression from the patch. That still must be inspected, classified, and disclosed before publication; it must not be silently discovered afterward.

## Root cause

The public-write guard in `prompt_text` is a disclosure and secret-safety policy. Raw GitHub mutations still execute through `bash`, with no structured PR-readiness preflight. The harness does not require the session to:

- Resolve the exact head SHA and inspect already-running or completed branch CI.
- Distinguish patch-caused failures from failures reproduced on the base branch.
- Reconcile PR behavior claims with the paths actually exercised by committed tests.
- Include verification evidence and known failures in the PR body.
- Fall back to a draft or stop when readiness is unresolved.

Local green checks therefore become an unjustified substitute for remote readiness, and a syntactically successful `gh pr create` is treated as completion of the publication step.

## Proposed fix

Add a guarded PR-publication preflight for non-draft creation. Before allowing the outbound write, the harness should:

1. Resolve the proposed head SHA.
2. Inspect CI already associated with that SHA and wait for in-progress head checks when appropriate.
[...]

Issue #843, opened 2026-09-10
Ctrl-V reports no clipboard image when an image is present


yxlyx:
## Problem
Ctrl-V can repeatedly report `no image on the clipboard` when an image is reported to be present, instead of attaching it.

## Confirmed code issue
`grabClipboardImage` in `src/vision_clipboard.zig` returns null both when no supported image is available and when clipboard extraction fails. The PNG/file-URL/PDF cascade suppresses subprocess errors; export and conversion failures also collapse into null. The caller therefore presents an empty-clipboard diagnosis it cannot establish.

The specific extraction failure behind this report is not yet confirmed.

## Expected fix
Distinguish an absent image from clipboard access, export, and conversion failures. Preserve privacy-safe failure categories for diagnosis and show an actionable error instead of asking the user to copy an image again. Investigate the false-negative extraction path and verify image attachment as well as genuinely empty-clipboard handling. Do not collect clipboard contents.

justrach:
Verified on the release branch: clipboard extraction distinguishes empty contents from access, export, conversion, timeout, and clipboard-change failures. Native pasteboard regression coverage exercises supported image formats, file URLs, invalid data, text, and empty contents. The release test suite passes.

Issue #840, opened 2026-09-10
Acknowledged peer handoff does not block later PR publication


yxlyx:
## Problem

Two co-resident sessions can explicitly agree that one session owns the remaining Git and GitHub publication work, yet the other session can still create the pull request. The shared-worktree checkpoint establishes awareness only; after it has fired once, neither an acknowledged handoff nor a promise to stop creates an enforced blocker.

This leaves the same branch, issue, commit, and pull request claimed and reported by multiple sessions even though only one public artifact exists.

## Reproduction

1. Start two sessions in the same repository and let both touch related work.
2. Have one session state that it owns push, pull-request creation, and check monitoring.
3. Have the other session acknowledge that handoff and state that it will perform no further Git/index/HEAD operations.
4. While the owner is still verifying, let the second session poll for a pull request.
5. Because no pull request exists yet, have the second session run `gh pr create` without an explicit release or handoff reversal.
6. The command succeeds, and both sessions can subsequently monitor or report the same pull request.

The absence of a pull request is incorrectly usable as an implicit timeout. Ownership never actually changed.

## CLI behavior is not the cause

The publication command succeeds normally. Later non-zero CLI results can be expected and unrelated:

- `gh pr checks --watch` exits non-zero when a check fails.
- A branch-protection REST query can return 404 when that endpoint is unavailable for the repository.

Neither result explains or justifies the duplicate ownership. Retrying or changing GitHub CLI commands cannot fix this race.

## Root cause

`presence.gateCheck` is deliberately a one-shot awareness checkpoint: seeing a live peer ACKs that peer, and reissuing the mutation proceeds. `peer_message` is an advisory mailbox and carries no structured, durable ownership state. There is therefore no atomic claim checked before local Git mutations or remote GitHub writes, and no requirement that the current owner explicitly release or hand off an artifact.

A session can consequently infer abandonment from polling, elapsed time, or visible progress and self-authorize a takeover. The harness has no state capable of rejecting that decision.

## Proposed fix

Add a repository-scoped artifact-claim protocol with explicit acquire, release, and handoff operations. Claims should cover at least branch, issue, commit/publication work, and pull request.

Before staging, committing, pushing, or performing a GitHub write, atomically check the relevant claim. A conflicting live claim must block the operation; acknowledging the ordinary shared-tree warning must not clear it. Time passing, an absent pull request, or repeated polling must not transfer ownership.

[...]

PR #852, opened 2026-09-10
Fix missing titles on existing GUI conversations

yxlyx:
## What changed
- Recognize both user-text content forms when saving session titles, skip empty and peer-injected content, and discard stale placeholders on resume.
- Recover legacy GUI titles consistently across listing, search, and opening a conversation. Invalidate cached recovery when the saved file changes.
- Add regression and edge-case coverage for persistence, malformed and image-only messages, Unicode boundaries, cache rewrites, and API consistency.

## Why
Existing conversations could be saved as "Untitled session" because the title fallback ignored structured user text. Correcting persistence prevents new occurrences; display-time recovery repairs existing entries without rewriting conversations or replacing named titles. Legacy reads remain bounded, and malformed or oversized files retain their saved fallback.

## Verification
Desktop unit tests and production build passed. The complete tier-1 pre-push checks passed.

Fixes #830.

PR #859, opened 2026-09-10
Merge release v0.0.296 into main with feature history preserved

### LINKED ISSUE #820: Line REPL image chips can split or repeat stale ordinals

### LINKED ISSUE #823: GUI Browser rejects ordinary search terms with Invalid URL
## Problem
The GUI Browser address field rejects ordinary words instead of searching. Enter a multi-word search query and submit it with Go; the pane shows:

`Error invoking remote method 'browser': TypeError: Invalid URL`

## Cause
Address handling treats non-scheme input as an HTTPS URL and passes it to the URL parser. There is no search-query fallback, and the parser failure is displayed directly.

## Requested resolution
Prefer browser-style address handling: navigate valid URLs and hostnames, and send ordinary search terms to a search engine with proper query encoding. Preserve URL safety restrictions rather than treating blocked schemes as search queries.

If search is intentionally unsupported, rename the Browser feature and label its input clearly as URL-only, with a friendly validation message instead of an IPC error. The current naming implies ordinary browser search functionality.

## Acceptance
- Ordinary words and multi-word queries produce search results, or the feature is clearly named and labeled as URL-only.
- Valid web addresses still navigate normally.
- Unsupported input does not expose a raw remote-method error.

### LINKED ISSUE #824: GUI settings: choose where web links open
## Problem
GUI web links currently open in the system default browser. There is no destination preference in the link handler, even though Graff includes a built-in Browser. Users should be able to keep linked pages inside Graff instead of switching apps.

## Requested change
Add a persistent setting in GUI Settings for where web links open:
- System default browser
- Graff built-in Browser

Route GUI web-link clicks through the selected destination. When Graff is selected, reveal its Browser and navigate to the clicked URL without replacing the main app view. When the system browser is selected, respect the operating system's default browser.

## Acceptance
- The preference is discoverable in GUI Settings and persists across restarts.
- Web links in GUI conversation content follow the selected destination consistently.
- Internal app navigation remains unchanged, and existing URL safety restrictions remain enforced.

This is separate from #823, which concerns searching from the Browser address field.

### LINKED ISSUE #825: Desktop: make Tasks sidebar optional and clarify the four-pane split limit
## Problem
Selecting a conversation with tasks automatically opens the right-hand Tasks sidebar, even when the user only wants to read or scroll through the conversation. There is no close control while remaining on that chat. The sidebar takes space away from already narrow split panes.

Cmd+D also silently stops adding panes after four.

## Reproduction
1. Use Cmd+D to reach four chat panes; press it again.
2. Select a conversation that has ongoing tasks.
3. Try to dismiss the Tasks sidebar without leaving the conversation.

## Cause
Task sidebar visibility is driven directly by the presence of tasks rather than an explicit user visibility preference. Its header has no dismissal control. The split action returns immediately at a hard-coded four-pane cap without feedback.

## Requested change
- Provide an explicit Tasks button to show ongoing tasks on demand, rather than opening the sidebar just because a conversation has tasks.
- Add a close control and preserve dismissal across task updates and conversation navigation until the user explicitly reopens it.
- Keep conversation panes readable when Tasks is open.
- Reconsider the four-pane cap; if retained for readability, make the limit discoverable and provide feedback instead of silently ignoring Cmd+D.

### LINKED ISSUE #826: Hide task expanders when there are no details
## Problem
Task rows display a disclosure chevron and toggle an expanded state even when there is nothing to show. This suggests task details are available when they are not.

## Cause
The task list supplies labels and statuses but no detail entries. The shared row component defaults missing details to an empty list, yet renders the disclosure button and expandable container unconditionally.

## Expected behavior
- Render rows without details as non-expandable: no chevron, expansion action, expanded-state accessibility attribute, or empty detail spacing.
- Preserve expansion for rows with actual detail entries.
- Determine expandability from available details, not task status; progressing or completing a task alone does not supply details.

### LINKED ISSUE #827: GUI chat should scroll only vertically, not sideways

### LINKED ISSUE #828: Folder picker can remain on Reading folder after Enter
Submitting a folder path with Enter can leave the folder picker displaying “Reading folder…” after its listing disappears.

Expected: finish browsing the submitted folder or display an actionable error.

Status: unresolved. The cause has not been identified. No fix has been implemented, and no pull request has been opened for this report.

### LINKED ISSUE #829: Desktop: Check for Updates shows a misleading installation warning
## Bug
Selecting **Codegraff → Check for Updates…** displays:

> Install a signed release in Applications to receive updates.

The update check does not proceed, and the message does not explain the actual reason updates are unavailable. **Automatically Download Updates** is also shown enabled and checked.

## Reproduce
1. Open the Codegraff application menu.
2. Click **Check for Updates…**.
3. Observe the installation warning instead of an update-check result.

## Cause and proposed fix
The update-menu action remains enabled when the updater is unavailable. All availability failures share this installation warning, although the availability gate checks packaging, platform, smoke mode, volume location, and update-configuration presence—not signing status. Which condition triggered this report is not yet established.

Track the specific unavailability reason and show accurate, actionable feedback. Reflect updater availability in both update controls, and ensure supported installed releases can perform a manual check.

### LINKED ISSUE #830: Existing GUI chats stay labeled Untitled session
Existing chats can remain labeled "Untitled session" after user messages have been saved.

The saved-title fallback recognizes text content blocks but not input_text blocks, so it persists the placeholder. Resuming then treats that placeholder as an explicit title.

Fix: recognize both text forms, skip empty and peer-injected content, discard the stale placeholder on resume, and recover a prompt-based display title for existing GUI saves without overwriting named titles.

### LINKED ISSUE #832: Automated GUI tests launch Electron windows and repeatedly steal desktop focus
## Problem
During Graff-driven GUI testing on macOS, Electron app windows unexpectedly appear in the foreground and take focus away from the application the user is working in. Switching back does not resolve it: continued testing repeatedly brings the test app forward, preventing normal use of the computer while tests run.

## Expected behavior
Automated GUI verification must not take over the user’s desktop. Default test runs should remain headless, hidden, or otherwise isolated from the active desktop, without activating windows or redirecting keyboard and mouse input. Any test that genuinely requires foreground interaction should require explicit opt-in rather than launching unexpectedly.

## Investigation / requested fix
Identify the test launch and window-activation paths responsible and make background-safe execution the default. Running a command as a background process is not sufficient if its GUI still activates. The exact triggering command and code-level cause are not yet confirmed.

## Acceptance criteria
- Start automated GUI tests while another application is focused: that application stays focused.
- Switching to another application during testing is respected; subsequent test steps and window launches do not steal focus back.
- Default test runs do not leave visible test windows behind.
- Foreground-only checks are explicitly opt-in.

### LINKED ISSUE #834: SDK packed-install smoke can resolve a released binary instead of its fixture

### LINKED ISSUE #835: Linux subagent cancellation test iterates a directory without iteration enabled

### LINKED ISSUE #836: Isolate TUI probe clipboard commands
## Problem
Tuiguard probes can interfere through the system clipboard, producing false selection failures. Serializing one pool does not prevent interference from other test processes or applications.

## Cause
Clipboard-using probes and their Graff children share the host clipboard.

## Fix
Give each offline gate probe private clipboard commands inherited by its child. Keep selection and OSC52 assertions intact and independent probes parallel. Direct probe runs retain native clipboard integration.

### LINKED ISSUE #837: Desktop composer skill-selection regression fails under pointer input

### LINKED ISSUE #839: GUI shows an active REPL session as finished
## Problem
Opening a session in the GUI while it is still working in the REPL shows an idle-looking conversation and a normal follow-up composer. Intermediate assistant commentary appears finished even though execution continues in the terminal.

## Reproduce
1. Start a multi-step task in the REPL.
2. While it is still running, open the same saved conversation in the GUI.
3. Observe that the GUI looks finished rather than indicating ongoing external activity or snapshot-only status.

## Cause
The saved-session transcript projection unconditionally marks assistant turns as done. The saved-session response carries no live execution state, and the GUI composer derives busy state from its own local activity tracking. Shared history is therefore mistaken for completed execution.

## Expected / proposed fix
Make the cross-surface contract explicit: either attach to live session activity and reflect its running/completed state, or clearly identify the view as a saved snapshot with unknown live status. Do not infer completion from intermediate commentary or a successful snapshot load. Make continuation ownership clear before accepting work on a session still running elsewhere.

Validate by opening an active REPL session in the GUI and checking both its in-progress display and its eventual completion or snapshot labeling.

@justrach

### LINKED ISSUE #841: Desktop updates disabled by the packaged executable name
Signed desktop builds can remain on an older release without automatically checking for updates.

The bundle retains the default Electron executable name. Electron consequently reports app.isPackaged as false, and the desktop updater treats the release build as unavailable.

Fix: rename the main executable during packaging, update CFBundleExecutable before signing, and reject distribution builds that retain the development executable name. Assert packaged status in the launch smoke check.

### LINKED ISSUE #842: Deferred shortcut focus can override newer chat navigation
## Problem
A delayed shortcut focus callback can move focus back to an older chat after a newer split has become active.

## Cause
The animation-frame callback focuses its captured chat without checking whether that chat is still active.

## Fix
Check the current active chat before applying deferred focus. Cover an older queued focus frame delivered after opening a new split.

### LINKED ISSUE #844: Completion can drop unresolved verification and report success
## Problem
An agent can replace its remaining verification item with a completed summary, then close the goal despite unresolved checks. The run can also be classified as successful without verified task success.

## Cause
`clearEpochForReplace` removes omitted open items. `applyTodoWrite` reports the dropped count but marks the replacement as fresh completion evidence. `completionGate` checks the resulting list without retaining the abandoned obligation.

Separately, `mainloop_trace.record` derives `Outcome.success` from a normal agent return, conflating execution success with task verification.

## Fix
Preserve unresolved acceptance requirements across checklist replacement unless the user changes scope. Dropping work must not satisfy completion. Track execution status, claimed completion, and verified outcome separately; an ordinary return is not proof of verification.

Add regressions for replacing the last open verification item with a completed summary, completion with unresolved validation, and normal returns that must not produce verified-success telemetry.

### LINKED ISSUE #845: TUI: queuing a follow-up freezes the thinking indicator
## Problem
While a turn is thinking, typing a queued follow-up freezes the thinking indicator. Queue composition should not make the active turn appear stopped.

## Cause
The line REPL sets g_steer_visible when the steering input appears. spinnerTask skips every animation frame while that flag is set, protecting the input row by freezing the activity indicator. This establishes a rendering pause, not cancellation of the underlying model request.

## Expected fix
Keep the thinking indicator live while composing and queuing follow-ups, without overwriting the input row. Ordinary queue input must leave the active turn running; interruption must remain an explicit action.

## Acceptance
During an active thinking turn, type and submit a queued follow-up. The thinking indicator must continue updating, the input must remain intact, and the follow-up must wait until the active turn finishes.

### LINKED ISSUE #846: Fresh-connection transport tests reject valid warm-up anchors

### LINKED ISSUE #848: Transport replay fixtures reject valid fresh prewarm anchors

### LINKED ISSUE #849: MCP connection-reuse test can hang during teardown

### LINKED ISSUE #850: Explicit bash timeout bypasses foreground auto-background safety

### LINKED ISSUE #851: Shared-tree guard checks the caller instead of the explicit Git target
## Problem
A mutating command such as `git -C <isolated-worktree> cherry-pick <commit>` can receive the shared-tree ownership checkpoint for the caller’s checkout, even though it targets a separate worktree. The action is not executed and must be resubmitted.

## Cause
`isSharedTreeGit` skips the value of `-C` and returns only a mutation classification. `gateCheck` then checks the global current-worktree identity, without receiving the command’s effective target. A shell `cd` before Git has the same mismatch.

## Fix
Resolve the effective mutation target before checking ownership, including explicit Git directory options and safely resolvable shell directory changes. Keep conservative handling for ambiguous commands.

Regressions should prove that an unrelated checkout owner does not block an isolated target, an owner of the actual target still triggers the checkpoint, and a subdirectory of the same worktree does not bypass it.

justrach:
Merge the v0.0.296 release and original feature branches into main using merge commits. Preserve the original feature commits and authors alongside the release integration history.

Includes desktop navigation and link settings, optional Tasks, workspace Agents, saved-session handling, interrupted-turn recovery, hidden GUI verification, native macOS CI, bounded shell waits, and regression repairs. Reconciles duplicate ADR numbers while retaining both decisions.

## Verification

The combined head passed the required pre-push checks and GitHub Linux, Windows, SDK, desktop, and native macOS GUI checks. The production app passed packaged launch and hidden-window focus verification.

The merge must use a merge commit; do not squash or rebase.

Closes #820.
Closes #823.
Closes #824.
Closes #825.
Closes #826.
Closes #827.
Closes #828.
Closes #829.
Closes #830.
Closes #832.
Closes #834.
Closes #835.
Closes #836.
Closes #837.
Closes #839.
Closes #841.
Closes #842.
Closes #844.
Closes #845.
Closes #846.
Closes #848.
Closes #849.
Closes #850.
Closes #851.

Refs #817, #840, #843, #847. These remain open for their outstanding acceptance or verification work.


justrach:
## CI follow-up (`12d5402`)

2 of 7 were red on `1a3043d` after the packed-install pin. Neither was a revert of #852 / #836 / #835 / #834.

### zig — Codex WS chain re-anchor
**Root cause:** `closeCodexWs()` zeros `codex_sent_upto`, so the rebuild's fresh socket is `eligible()` for `generate:false` prewarm and the full-input request chains onto `resp_prewarm_N`. The PTY assertion required *no* `previous_response_id` at all, so a valid fresh-socket prewarm parent failed as a stale chain.

**Fix:** record each mock connection's observed prewarm id and require `has_fresh_parent` (this socket's prewarm or none — never `resp_chain_1` or another socket). Same contract as 296 `9a0ab8e`.

**Checked locally:** `python3 scripts/test-pty-codex-ws.py zig-out/bin/graff` — chain re-anchor + mid-turn compaction + transactional rollback all green.

### desktop — `test:projects` composer interactions
**Root cause:** one-shot pointer coords were measured, then hover scrolled/repositioned the portaled "Add photos" row, so `fileChooserClicks` stayed 0.

**Fix:** settle geometry against `elementFromPoint` before press; park the pointer before IME checks; keep a stale Cmd+1 focus frame from stealing a new split. Same as 296 `788647b` / `f49c534`.

**Not verified here:** Electron visual suite is macOS CI only.

### Left intact
#852 wrap fixtures, #836 private clipboard + quiet-stdin xclip, #835 iterate dir, #834 packed-install pin. No force-push of yxlyx branches. No merge.

Tier 1 green on push (2131 tests, 19/19 tuiguard).

justrach:
## CI follow-up (`1abe543`)

Desktop `test:projects` was still red on `12d5402`. Pointer settle and the visual-test race harness were already in place; they were not the miss.

### Exact failure
`runComposerInteractions` in `apps/native/electron/composer-interaction-visual.cjs:115`

```
AssertionError: An older shortcut focus frame must not steal the new split
1 !== 3
```

GitHub Actions run `34482704974` / job `102889190989` (`bun run test:projects`).

### Root cause
`12d5402` added the delayed Cmd+1 RAF *test* (same as 296 `f49c534`) but left production `useDesktopShortcuts` focusing the old textarea unconditionally. That `.focus()` hits `onFocusCapture` and `focusChat(1)` steals `activeId` from the new split (`3`).

### Fix
Only apply the deferred textarea focus while `ref.current.activeId === id` — same production guard as 296 `f49c534` / #842. No change to #852 wrap, #836 xclip poll, #835 iterate, #834 packed-install, or the prewarm parent assert.

**Not verified here:** Electron visual suite is macOS CI only.

PR #838, opened 2026-09-10
GUI: choose where web links open

yxlyx:
## What changed
- Add a persistent Settings choice between the system default browser and Graff's built-in Browser.
- Route normal and new-window links through shared safety checks and reveal the focused chat's Browser without replacing the app.
- Add persistence, URL edge-case, and Electron interaction coverage.

## Why
Conversation links previously always left the app. Desktop-backed storage survives restarts and origin changes, while shared interception keeps link behavior consistent without coupling individual renderers to the destination preference. System-browser behavior remains the default, and internal navigation stays separate.

## Failures uncovered during verification
- Ignore stale deferred shortcut focus after newer navigation (#842).
- Match packed SDK mock versions to dependency pins (#834).
- Enable directory iteration in the Linux activity test (#835).
- Give gate probes isolated clipboard commands instead of sharing host state (#836). Direct probe runs retain OS integration.
- Recognize only the fresh connection's prewarm anchor in transport fixtures, retaining full-history and stale-anchor checks (#848).

## Validation
Desktop, SDK, Windows, and Zig GitHub checks pass on the final revision. Local desktop unit tests, production build, project/link interaction tests, SDK packed-install smoke, pool integrity tests, transport scenarios, and the full offline tier-1 hook also passed. The hook passed with its normal parallel pool and no skipped checks.

Closes #824

PR #861, opened 2026-09-10
fix: shared-tree gate checks the git target, not the caller (#851)

### LINKED ISSUE #851: Shared-tree guard checks the caller instead of the explicit Git target

justrach:
Leftover from the open-issue wave that is still broken on `main` and is not in #859 or the open yxlyx feature PRs.

## #851

`isSharedTreeGit` already skipped `-C` when classifying the subcommand, but `gateCheck` then compared the caller checkout. `git -C <isolated-worktree> cherry-pick` (or a leading `cd` into that tree) hit the shared-tree checkpoint for the wrong tree.

Resolve the command target first. A different worktree identity skips the caller checkpoint; a subdirectory of the same tree still gates; quoted or globbed paths stay conservative.

## Verification

- `mutationTarget` parses `git -C`, glued `-Cpath`, and a leading `cd`.
- `identityAt` distinguishes two `git init` checkouts and matches a subdirectory of the first.
- Isolated `git -C` does not apply the caller gate; same-tree `-C` / `cd` and bare `git cherry-pick` still do.

Does not include #835 (cancel-file iterate) or the other wave fixes already on #859 / #821 / #822 / #831 / #833 / #838 / release 0.0.296.

Fixes #851.

### DIFF
--- a/src/agent_tool_gate.zig
+++ b/src/agent_tool_gate.zig
@@ -220,10 +220,12 @@ pub fn gateTool(self: *Agent, call: ToolCall) !?ExecResult {
         // approvals prompt would ever surface the collision. The model sees
         // the peer's identity + goal and re-issues the command to proceed.
         if (presence.isSharedTreeGit(cmd) or presence.isSharedTreeShell(cmd)) {
-            if (presence.gateCheck(self.io, self.arena)) |checkpoint| return .{
-                .text = checkpoint,
-                .is_error = true,
-            };
+            if (presence.sharedTreeGateApplies(self.gpa, self.io, self.arena, cmd)) {
+                if (presence.gateCheck(self.io, self.arena)) |checkpoint| return .{
+                    .text = checkpoint,
+                    .is_error = true,
+                };
+            }
         }
         const destructive_git = Approvals.isDestructiveGit(cmd);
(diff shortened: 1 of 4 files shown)


justrach:
Folded into #859 with `git merge --no-ff` (no squash). The #851 commit `c8afe75` is an ancestor of #859 HEAD `e4c275e`.

Closing this leftover draft: it is based on `main`, where Zig CI already crashes on #835 (Linux cancel-file BADF). That fix is already on #859, and a main-based leftover cannot go green independently. Fewer PRs; the work lives on https://github.com/justrach/codegraff/pull/859.

PR #965, opened 2026-09-15
fix: report downsized workflow tasks

### LINKED ISSUE #863: workflow silently executes only the first task in a multi-task phase
## Observed
A workflow invocation supplied two dependent phases: a sweep with four independent read-only tasks, followed by one synthesis task consuming the previous results.

Only the first sweep task ran. The returned summary reported:

```text
phase 1/2 sweep: 1/1 ok, 0 retried
phase 2/2 synthesize: 1/1 ok, 0 retried
```

The synthesis input contained only the first labeled research result, and the synthesis explicitly reported that the other three inputs were missing. There was no warning that tasks had been omitted. A local event trace recorded one successful workflow invocation lasting about 134 seconds.

## Expected
All four requested tasks should execute, queuing if necessary, before synthesis. If a task limit or execution policy prevents that, return an explicit warning or failure instead of silently shrinking the phase to one task.

## Reproduction shape
Supply phases with task counts of four and one respectively, with the second phase consuming the previous results. Compare requested task counts against execution summaries and the synthesis input.

The cause is not established. This report intentionally excludes repository content, task prompts, local paths, and session identifiers.

yxlyx:
Closes #863

## What changed

- Report `requested`, `executed`, and `omitted` counts whenever workflow admission limits a phase width.
- Carry that warning into downstream synthesis, all-failed abort evidence, and the final workflow manifest.
- Keep conditional `when` gates matched against worker evidence only, so warning metadata cannot open a gated phase.
- Add unit and scripted behavior coverage for successful, gated, and all-failed downsizing.

## Why

Workflow admission intentionally reduces some fan-out plans to fit orchestration policy and call budgets. The executor previously sliced the authored task list before reporting counts, so a four-task phase could appear as a complete `1/1` phase and synthesis received partial evidence without an explicit warning.

This approach preserves the existing admission policy and worker counts while making every omission visible at the points that consume or report the result. The warning is kept separate from gate evidence to avoid changing workflow control flow.

Removing downsizing entirely was rejected because it would discard the existing budget-preserving policy. Returning an error for every downsize was also rejected because the policy deliberately permits useful partial fan-out; the required fix is honest accounting rather than changing admission semantics.

## Verification

- `scripts/eval-tier1.sh` — passed.
- `python3 scripts/eval-tier2.py --only escalation-research-ask-gets-one-scout` — passed.
- `python3 scripts/eval-tier2.py --only workflow-downsize-warning-does-not-open-gate` — passed.
- `python3 scripts/eval-tier2.py --only r0d-advisory-carries-parked-abort-evidence` — passed.
- GitHub `ci` push checks on `0eb1bb7a4abab40f6fb12e088c3c7215ce5e549a`: `zig`, `windows`, and `sdk` passed.


justrach:
Already on release/v0.0.298 as 888ef6a1.

Issue #1001, opened 2026-09-17
Peer DMs should start a turn so sessions can reply without a human relay


justrach:
## Symptom

`peer_message` delivery is pull-only at the receiver's next user step. An idle session parks the body and paints a `[peer]` wake, then waits. The human has to send another message (or paste the ping) before the model can `action=inbox` and reply. Two live sessions cannot finish a back-and-forth on their own.

Related: #469 (channel), #563 (speech; closed). This is the leftover: hear-on-idle, not the history firehose #563 already fixed.

## Cause

`deliverInbound` runs at the start of `runTurn` only (`src/agent.zig`). It parks bodies in `peer_inbox.zig` and injects a one-line wake (ADR 0004). Idle roots never call `runTurn`, so they never drain or answer. Children never drain (`root.sub` returns). A standing liaison is rejected on purpose.

Posted ≠ heard. The sender's tool result is "posted"; the receiver does not start a turn.

## Fix

When a live **idle root** receives addressed inbound, start a model turn on the existing wake + `action=inbox` path. Keep going until that turn ends (the model finishes, or it does not send another `peer_message`). No user message required.

Keep ADR 0004: history stays a one-line wake, bodies stay in the inbox ring, do not stuff the room into durable chat, do not spawn a standing liaison.

Bound it: no infinite ping-pong (cap cascaded turns per pair / require an explicit stop). Mid-turn preemption stays out of scope (#430).

justrach:
Idle TUI auto-turn is in #1003 (JSONL stays the durable room; Accord Unix is opt-in live). Line-REPL/ACP idle is still open.

justrach:
Shipped in v0.0.299: idle TUI and line REPL start a turn on new peer mail.

PR #1003, opened 2026-09-17
Opt-in Accord live path and idle TUI peer wake (#1001)

justrach:
JSONL stays the durable peer room. GRAFF_ACCORD=1 may also send the same line over Unix 0600 (default off). Idle TUI starts one turn per new unread batch (#1001). Rebased onto latest release/v0.0.299. Local: pre-push tier 1 green. Remote: CI not yet. Does not close #1001 for line-REPL/ACP.


justrach:
Landed on release/v0.0.299 (fast-forward). Closing the draft instead of merging to main.

PR #1105, opened 2026-09-20
test(native): cover repeated folder picker submissions

### LINKED ISSUE #1089: Add regression coverage for repeated folder-picker submissions
Repeated submissions in Open a folder previously caused the current directory listing to disappear while navigation was pending. The runtime behavior is fixed, but the desktop interaction suite does not lock it in.

Expected: repeated Enter submissions keep the current listing visible until the newest request settles.

Add a deterministic interaction regression using delayed folder responses.

yxlyx:
Closes #1089

## What changed

- Add a focused `test:folder-picker` visual-suite entry that reaches the rendered folder dialog through the navigation runner.
- Send three desktop Enter key submissions through the named test-input helper and hold their folder-listing requests pending.
- Settle each stale request separately and assert after each one that the current listing remains visible while the newest request keeps navigation busy.
- Monitor every rendered frame while the requests settle and fail if the `apps` row is absent, CSS-hidden, transparent, zero-sized, or outside an overflow-clipping rectangle between the row and the dialog.
- Settle the newest request and assert the dialog re-enables folder selection with the listing intact.

## Why

Repeated Enter submissions issue overlapping navigation requests. The runtime fix is already on `main`, but the interaction suite did not exercise that asynchronous overlap, so a later change could reintroduce a blank or flickering dialog body unnoticed. This uses the real rendered dialog, the existing desktop input dispatcher, and deterministic delayed responses without changing runtime behavior.

## Verification

Local: `cd apps/native && bun run test:folder-picker` — passed.

Remote: exact-head CI passed.


justrach:
Landed on the 0.0.302.2 line.

PR #1108, opened 2026-09-20
Fix native composer history recall layout

### LINKED ISSUE #1106: Native composer reshapes during prompt history recall in split chats
## Symptom

Pressing Up to recall a previous prompt in a split native chat visibly reshapes the composer: the textarea changes width and the controls move to another row.

## Cause

The compact live composer switches between inline and stacked grids based on the recalled prompt's measured width.

## Fix

Keep live composers on a stable stacked grid and add an Electron interaction regression that verifies history recall preserves the textarea width and control placement.

yxlyx:
Fixes #1106

## What changed

- Keep live native composers on a stable stacked grid so prompt history recall does not relocate controls or change the textarea width in split chats.
- Add an Electron interaction regression that recalls a previous prompt with ArrowUp and verifies the compact composer remains stacked with stable width.

### Problem / failure mode

A compact live composer selected its inline or stacked grid from the current draft width. Recalling a medium-length prompt changed that decision in one paint, so the textarea widened and the controls moved to another row, making the whole prompt card visibly reshape.

### Reason for this approach

The layout should be stable before and after recall. Keeping live composers stacked removes the content-dependent grid transition while retaining the adaptive behavior for component demos.

Split chats use a little more vertical space for short drafts, in exchange for stable geometry while recalling and editing prompts.

### Rejected alternatives

Animating or delaying the layout switch would only mask the reflow; preserving the inline grid for recalled text would leave long prompts cramped beside the controls.

## Verification

Local: `cd apps/native && bun run build` — passed, including TypeScript.

Local: `cd apps/native && GRAFF_VISUAL_SUITE=interactions bun scripts/test-visual.mjs` — passed, including keyboard navigation, composer interactions, file recovery, and chat scrolling.

Local: `scripts/eval-tier1.sh --only reach,build` — blocked by the existing Zig 0.16 `addPassthruArgs` incompatibility; the same failure reproduces unchanged on `origin/main`.

Remote: exact head `0f62056237879e7e7960c1b69dc213b000772939` — `sdk` and `windows` passed. `zig` failed in `scripts/regression-release-297.py`, matching current `main` SHA `772d683b16fb8adf8931b4680a65bcb3ee1c18a5`; `desktop` and the foreground portion of `native macOS GUI` failed on the same folder-dialog timeout as current `main`. The macOS workflow also reported an unrelated retained-server shutdown fixture failure. This PR remains draft while checks are red.


justrach:
Landed on the 0.0.302.2 line.

PR #1113, opened 2026-09-20
test(native): cover retained worktree path wrapping

### LINKED ISSUE #1110: Cover retained worktree paths in native chat overflow tests
The native chat overflow suite covers long prose, URLs, inline paths, errors, recaps, reasoning, code and tables, but it does not exercise the plain retained-worktree notice shape.

Add a long `.worktrees/...` notice to the fixture and assert that it wraps onto multiple in-bounds lines across narrow and wide transcript panes.

yxlyx:
## What changed

- Add a long plain-text `.worktrees/...` retained-worktree notice to the native chat overflow fixture.
- Assert in both transcript panes that the notice wraps onto multiple lines and every line remains inside the transcript at 260, 360, and 720 px widths.

## Why

**Problem/failure mode:** the existing overflow suite covers generic prose and inline paths, but not the retained-worktree notice emitted by isolated runs. That leaves a real long-path shape unprotected from future horizontal-overflow regressions.

**Reason for this approach:** extend the existing fixture and its range-based geometry assertion, so the test follows the same production rendering path and proves glyphs remain readable rather than merely hidden.

**Constraints/trade-offs:** this is regression coverage only; the general wrapping implementation is already on `main`. Wide code blocks and tables must keep their own local horizontal scrolling, so the new assertion targets the prose paragraph rather than banning all horizontal overflow inside the transcript.

**Rejected alternatives:** duplicating the production CSS or adding another fixture route would add behavior or test surface without improving coverage. A synthetic path avoids machine-specific data while preserving the `.worktrees/...` shape.

Fixes #1110

## Verification

- Local: `cd apps/native && bun run build && bun run test:overflow` — passed.
- Local tier 1: `scripts/eval-tier1.sh --only reach,build` — fails identically on this head and clean `origin/main` because the local Zig 0.16 toolchain rejects the vendored `addPassthruArgs` call before project tests run.
- Remote: CI pending.


justrach:
Landed on the 0.0.302.2 line.

Issue #1099, opened 2026-09-20
Native GUI attachment timeout can block multi-file uploads without recovery


yxlyx:
## Symptom

The native GUI can remain in an attachment-progress state while also reporting:

> Adding the file timed out. Try again.

With multiple selected files, the message does not identify which file failed or whether the remaining files are still pending. Sending remains unavailable while the shared upload counter is nonzero, and the UI provides no per-file retry or cancellation path.

## Failure mode

The composer currently increments one shared upload count, then awaits files serially. Each file uses an unbounded `fetch` to the attachment endpoint, while all failures are collapsed into one shared error string.

Consequences:

- A request that never settles can block every later file indefinitely.
- A partial multi-file failure can show an error while other work remains pending.
- The error lacks the filename and failed stage, so the user cannot identify the problematic file.
- A lower-level timeout is surfaced without an application-owned recovery path.
- Successful files and failed files are not represented independently, making safe retry unclear.

## Expected behavior

Attachment work should be bounded and tracked per file. One stalled file must not leave the composer permanently blocked or prevent unrelated files from completing.

## Acceptance criteria

- Give each attachment request an application-owned timeout and abort path.
- Track pending, successful, and failed state per file rather than only through one global counter and error.
- Include the filename in actionable failure text.
- Allow a failed file to be retried or removed without discarding successful attachments.
- Ensure the composer exits its progress state after every request settles, aborts, or times out.
- Ensure starting a new attempt cannot retain an unrelated prior error.
- Add a regression where the first request never resolves and a later file can still complete or fail independently.
- Add a regression for partial multi-file failure and verify that progress, send availability, and displayed errors recover correctly.

cc @justrach

justrach:
Fixed on `release/v0.0.302.2` (`6b23ae9b`).

PR #1125, opened 2026-09-20
Fix multi-file attach timeouts in the native composer

justrach:
Closes #1099

## What changed / Why

A hung `/api/attach` request blocked every later file because the composer uploaded sequentially, and a failure dropped to a shared error with no filename or retry. Each file now uploads in parallel with its own timeout, the error names the file, and a failed chip can be retried or removed without dropping files that already landed.

Local: `bun test lib/attachments.test.ts` — 15 pass


justrach:
Landed on `release/v0.0.302.2` (`6b23ae9b`).

PR #1127, opened 2026-09-20
Keep accepted Steer now pending until the turn ends

### LINKED ISSUE #1101: Native GUI: accepted queue steer reports a false interruption failure after 10 seconds
## Problem

In the native GUI, steering a queued follow-up can show this error even after the cancellation request was accepted:

> Could not confirm interruption. Your selected message is still next; retry Steer now or wait for this turn to finish.

Reproduction:
1. Start a turn that takes more than a few seconds to stop.
2. Queue a follow-up.
3. Choose **Steer now**.
4. Wait at least 10 seconds while the original turn is still completing its shutdown.

The queued message remains selected, but the UI changes from "Steering…" to a failure. Repeating the action while the first attempt is pending is ignored.

## Root cause

The queue steerer starts a fixed 10-second confirmation timer. Sending `session/cancel` only acknowledges that the notification was written; the timer is cleared later, when the active prompt stream emits its terminal event or exits. A valid cancellation whose stream teardown takes longer than 10 seconds is therefore reported as a failure. The client cancellation helper also suppresses transport errors, so request delivery and turn completion are not represented as distinct states.

## Fix

Treat accepted cancellation and completed interruption as separate states. Do not convert a successfully accepted steer into an error solely because prompt-stream teardown exceeds 10 seconds. Keep the queued item selected until the current turn reaches its terminal event, report actual request/transport rejection, and make repeated steering input idempotent while cancellation is already in flight.

justrach:
Fixes #1101.

**Symptom:** an accepted Steer now flipped to a false failure after 10s.

**Fix:** an accepted `session/cancel` clears the timer and stays pending until the turn finishes. Timeout only fires if the request never settles. Transport rejection still errors. A repeat steer while in flight is a no-op.

Queue steerer tests cover the accepted, timeout, reject, and in-flight cases.

<div><a href="https://cursor.com/agents/bc-3a4fe7ff-cf82-4498-b368-22a99ed5bcfd?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3a4fe7ff-cf82-4498-b368-22a99ed5bcfd&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


justrach:
Landed on the 0.0.302.2 line.

PR #1129, opened 2026-09-20
Isolate concurrent agents onto their own Git worktrees

justrach:
Two live sessions in one checkout share `.git/index` and fail on `index.lock`. Independent fan-out also defaulted to that shared tree.

A second session in a claimed checkout now mints its own worktree. Independent children default to `isolation: worktree`. Pipeline stages stay on `shared_cwd` so later stages see earlier edits. Create mints from the remote base. Archive keeps dirty or unique-commit trees.

`git worktree add` failure is an isolation failure unless fallback is explicit.

Claim fixtures now read the shared Git-common-dir ledger, and a failed PR lookup stays unknown instead of releasing a foreign claim.


justrach:
Landed on the 0.0.302.2 line.

PR #1130, opened 2026-09-21
Show ask_user success only after delivery

### LINKED ISSUE #1114: Native ask_user can show Answers sent without delivery acknowledgement
## Summary

The native `ask_user` card can display **Answers sent** without confirming that the selected value reached the waiting ACP call. If delivery fails or an empty value is submitted, the session can remain blocked, repeat the decision, or appear inactive even though the UI claims the answer was sent.

This is distinct from #1087 / #1100. That fix removes the stale radio auto-submit race and adds a repeated-activation guard, but the success state is still optimistic and the ACP route can acknowledge a failed notification.

## Current behavior

The production flow has no end-to-end acknowledgement for the matching `call_id`:

1. `ApprovalCard.send()` calls `setSent(true)` before invoking `onSubmitted`.
2. `onSubmitted` is typed as a synchronous callback, so the card cannot wait for delivery or display an error.
3. The native harness starts `answer(...)` with `void`, discarding the promise and any rejection.
4. The ACP `session/answer` route catches a transport notification failure and still returns `{ "ok": true }`.
5. A non-cancelled empty answer is not rejected before it reaches the session.

The user can therefore see **Answers sent** when the ACP process never received a usable answer. The visible card no longer offers a retry, while the backend may still be waiting on the original question.

## Trajectory evidence

A saved native trajectory contained six `ask_user` calls. Three corresponding tool results were empty. The first empty result caused a near-identical publication question to be presented again. Near the end, two semantically equivalent draft-acceptance questions also received empty results; after the final one, the transcript ended without a completed model response or terminal turn event.

The same run also showed a long period with no model or tool event after an approval prompt, followed by a resumed model call only after another answer attempt. This makes the optimistic **Answers sent** state actively misleading: it cannot distinguish accepted, undelivered, empty, or stalled replies.

## Reproduction

1. Start a native session that emits a single-choice `ask_user` call.
2. Select an option and submit it.
3. Make the ACP transport unavailable before `session/answer` is notified, or submit through a path that produces an empty value.
4. Observe that the card immediately displays **Answers sent**.
5. Observe that the route can return success even when notification throws, while the session remains waiting or later asks again.

## Expected behavior

- Show **Sending…** while delivering the response.
- Show **Answer sent** only after the server confirms that the matching `call_id` was delivered to the live ACP session.
- Return a non-success response when the transport notification fails or the target call/session is no longer valid.
- Reject a non-cancelled empty answer.
- On failure, preserve the selected value and show a retryable error instead of replacing the controls with a success badge.
- Make submissions idempotent per `call_id`, including repeated Enter/click activation.
- Once an answer is acknowledged, clear the pending ask state and show subsequent thinking, text, and tool activity normally.

## Regression coverage

Add production-path coverage for:

- successful answer acknowledgement for the matching `call_id`;
- dead transport / notify failure;
- empty non-cancelled answer rejection;
- retry after failed delivery without losing the selected value;
- repeated Enter/click producing one delivery;
- the UI never showing **Answers sent** before acknowledgement.

Observed in the native desktop flow on Codegraff `0.0.301`. The remaining optimistic acknowledgement behavior is also present after the #1100 change that fixed #1087.

justrach:
Fixes #1114.

**Symptom:** the approval card could show Answers sent before the matching call reached the live session.

**Fix:** Sending… until the route confirms the call_id. Empty answers and notify failures stay retryable. The engine rejects empty non-cancelled session/answer the same way.


justrach:
Landed on the 0.0.302.2 line.

PR #1131, opened 2026-09-21
Use the composer radius on short prompt bars

### LINKED ISSUE #1090: Reduce excessive corner radius on text bubbles and inputs
## Problem

Text bubbles and text-entry surfaces use an overly large, pill-like corner radius. On short content, the curved ends sit too close to the text and make the UI feel cramped and visually heavy. This is visible in sent-message bubbles, the main composer, and text inputs inside choice dialogs.

## Expected

Use a smaller, consistent corner radius so short text has comfortable visual clearance from the curved edges while the surfaces still read as rounded containers.

## Scope

Review the shared styling for chat bubbles and text-entry containers so the adjustment is consistent across these surfaces.

justrach:
Fixes #1090.

**Symptom:** short prompt bars used a pill radius that pinched text.

**Fix:** the prompt shell always uses the composer token. User bubbles stay card.


justrach:
Landed on the 0.0.302.2 line.

PR #1132, opened 2026-09-21
Record why an ACP slot is terminated

### LINKED ISSUE #1107: Native desktop terminates active ACP turn with SIGTERM at one-hour boundary
## Symptom

A long-running native desktop turn is interrupted at roughly the one-hour boundary with:

```text
Response interrupted
After 1h 1m. Completed tool results are kept. Send a follow-up to continue.
graff acp exited (SIGTERM)
```

The turn had completed many model/tool cycles and was still making progress while a background command waited for remote CI. The child was not idle or stuck.

## Trajectory evidence

- The ACP run lasted 3,635,785 ms (60m 35.8s).
- It recorded 298 successful model calls and 479 completed tool calls.
- The final event was a successful model response; there was no normal turn/recipe outcome afterward.
- There was no recorded user cancellation, transport retry, stream-stall budget event, or tool error at termination.
- The persisted transcript ends immediately after starting a background CI watcher, so the remaining publication step was lost even though completed work and tool results were preserved.
- The UI reported that the ACP child exited under `SIGTERM`, not that the model/API failed.

## Relevant lifecycle behavior

The native ACP route gives `session/prompt` a 24-hour RPC timeout, so the visible one-hour boundary does not come from that request timeout. The backend sends `SIGTERM` through its slot cleanup path after a page/chat slot is disposed or replaced, but the current trace does not record which lifecycle action requested termination.

This suggests one of two defects:

1. an uninstrumented one-hour supervisor/watchdog is terminating an active ACP child, or
2. a page/session lifecycle event is incorrectly disposing an active slot near that boundary.

The evidence confirms the unexpected active-turn termination, but does not distinguish those triggers yet.

## Expected behavior

An active turn should continue beyond one hour while it is producing model/tool activity. Page/session cleanup should not silently kill an active turn unless the user explicitly closes, resets, or cancels it. If termination is intentional, the trace should record the initiator and reason before signaling the child.

## Suggested regression and diagnostics

- Run a native desktop ACP turn past one hour while it periodically emits activity and waits on a background job; assert that the child remains alive and the turn can complete.
- Record every slot termination with the chat/page lifecycle action, whether a turn was active, child age, and signal reason.
- Add coverage that `pagehide`, tab disposal, session bootstrap/reset, and any supervisor timeout cannot terminate an unrelated or still-active slot without an explicit reason.

Observed on Codegraff `0.0.301` on September 20, 2026.

justrach:
Fixes #1107.

**Symptom:** an active desktop turn could exit under SIGTERM near one hour with no recorded reason.

**Fix:** every slot termination records the lifecycle action, whether a turn was active, child age, and signal. Idle park does not kill a busy slot. A restored pagehide does not reap the page.


justrach:
Landed on the 0.0.302.2 line.

PR #1133, opened 2026-09-21
Expire orphaned publication claims

### LINKED ISSUE #1097: Stale publication claim blocks unrelated Git writes
## Problem

The artifact guard can report `artifact claim held` for a publication claim and block unrelated Git writes even when the reported owner has no matching ledger claim and release reports that no claim exists.

## Cause

Claim status and the owner ledger can diverge, leaving an unreleaseable guard entry that still appears live to other sessions.

## Fix

Use the same repository-scoped record for status and release, repair or expire orphaned publication claims, and do not let an unowned stale claim block unrelated artifacts.

justrach:
Fixes #1097.

**Symptom:** a stale publication claim could block unrelated Git writes while status and release reported no claim.

**Fix:** status and release use the same repository-scoped publication record as the write gate. Dead-owner publication claims expire instead of blocking.


justrach:
Landed on the 0.0.302.2 line.

PR #1134, opened 2026-09-21
Count local checks in publication preflight

### LINKED ISSUE #1094: PR publication preflight can ignore completed check evidence
## Symptom

PR publication preflight can block `gh pr ready` after the exact head's branch CI and PR-specific workflow have passed, including the workflow step that executes the regression under review.

The review reports that `observed_local_checks` is empty and that no supplied caller establishes the regression as runnable, even when the check was run in the current task and the unchanged repository call chain reaches it.

## Cause

Claim review appears limited to changed source files plus a generic head-CI observation. It does not incorporate completed local check observations, the PR check rollup, or unchanged caller/configuration files needed to establish test reachability.

## Fix

Include current-task check results and exact-head PR checks in publication evidence, and resolve test reachability through unchanged repository callers/configuration before marking a claim unresolved.

justrach:
Fixes #1094.

**Symptom:** publication preflight could ignore completed local checks and unchanged callers.

**Fix:** compound check commands and eval-tier scripts count as local checks. Review includes unchanged callers/configuration and exact-head PR check output.


justrach:
Landed on the 0.0.302.2 line.

PR #1139, opened 2026-09-21
Stop invented read_file storms; keep native hosts native

### LINKED ISSUE #1016: Programmatic file reads fail despite native tool availability
The programmatic tool runner rejects a literal read_file call with “MCP tool schema is not loaded” while the native read_file tool is available and works directly. Resolve native functions before deferred MCP lookup so both paths use the same available tool catalog.

### LINKED ISSUE #1036: Desktop browser tool: click with text (no selector) fails with opaque "Script failed to execute"
**Symptom**

The desktop app's embedded-browser tool (`browser`) fails when `click` is called with a `text` value instead of a `selector`. The same element clicks fine when addressed by CSS selector.

Repro shape: on any page with a filter link, `click {text: "1 Open"}` errors; `click {selector: "a[href*='is%3Aopen']"}` on the same page succeeds.

**Error returned to the caller**

```
Script failed to execute, this normally means an error was thrown. Check the renderer console for the error.
```

This is a dead end in practice: the renderer console isn't reachable from the chat, and the message doesn't say what threw.

**Root cause**

In `apps/native/electron/browser-actions.cjs`, the `click`/`fill`/`select`/`hover` branch builds its injected script from `params.selector` only:

```js
const selector = JSON.stringify(String(params.selector || ''));
...
const prefix = `const el=document.querySelector(${selector});if(!el)throw Error('Element not found');...`;
```

`text` is consumed by `fill` (the value to type) and `find` (the find-in-page query), but never by `click`/`hover`. A click with `text` and no `selector` therefore injects `document.querySelector("")`, which throws `SyntaxError: '' is not a valid selector` in the renderer. `webContents.executeJavaScript` rejects with Electron's generic wrapper sentence, dropping the underlying renderer error, and the automation endpoint (`automation.cjs`) forwards only that generic text. The tool schema (`desktop-tools.cjs`) advertises `text` for the browser tool as a whole rather than per action, so callers reasonably expect click-by-text to exist.

**Suggested fix**

1. Either implement text-based resolution for `click`/`hover` (match elements surfaced by `snapshot` — `innerText` / `aria-label` / placeholder — and click the first visible match), or validate up front and fail with an explicit message such as `click requires selector; text is only supported for fill/find`.
2. When `executeJavaScript` rejects, include the underlying renderer error alongside the generic sentence so failures are diagnosable from chat, e.g. `Script failed to execute: SyntaxError: '' is not a valid selector`.

### LINKED ISSUE #1085: Clipboard paste reports generic clipboard read failed
Ctrl-V or `/paste` shows:

`clipboard read failed — retry Copy; use Edit > Paste for text or /image <path>`

**Why.** Image paste shells out to a JXA AppKit helper. Any helper non-zero exit that is not the Apple Event denial code, plus cancelled or spawn errors, is classified as a generic access failure. Empty boards and text-only boards have other messages; this line is the catch-all, so a helper crash or unexpected osascript exit looks the same.

**Fix.** Do not lump helper failures into one access line. Distinguish helper crash vs permission vs empty/text-only, and fall through to normal text paste when the board has no image.

### LINKED ISSUE #1116: read_file miss storm: guessed sequential paths are not stopped
## Symptom

A turn can spend dozens of `read_file` calls on files that do not exist. The names increment in lockstep (for example `docs/adr/0365-…` through `0397-…`) and reshuffle the same slug words. None of those records exist. One listing of the directory, or the ADR index, would have shown that.

## Harness text

```
read_file: … does not exist (paths are relative …
```

Each miss still lands as a failed tool card.

## Cause

`read_file` reports not-found and returns control to the model. There is no consecutive-miss stop for same-prefix guessed paths. Bounded lexical repetition (ADR 0065) applies to prose delivery, not tool-call path guessing. Directory listing is a separate tool (`codedb list_dir`) and is not injected after a miss.

## Fix

After N not-found reads under the same directory prefix, or an incrementing numbered filename pattern, refuse further guesses in that turn and point at listing the directory. Do not keep executing invented paths.

justrach:
Fixes #1116. Fixes #1016. Fixes #1085. Fixes #1036.

## Symptom

A turn can spend many `read_file` calls on files that do not exist. Names increment under one directory prefix. Each miss is still a failed tool card.

A literal `read_file` host call can fail with `MCP tool schema is not loaded` while the native tool is available.

A clipboard helper non-zero exit is reported as a generic access failure.

Desktop `browser` click with `text` and no `selector` fails with an opaque script error.

## Harness text

```
read_file: … does not exist (paths are relative …
rlm: MCP tool schema is not loaded — call load_tool_schemas first
clipboard read failed — retry Copy; use Edit > Paste for text or /image <path>
Script failed to execute, this normally means an error was thrown.
```

## Cause

`read_file` returned not-found and gave control back. There was no consecutive-miss stop for same-prefix or incrementing numbered names.

RLM short-name resolve mapped `read_file` onto an unloaded MCP tool.

Helper crash, cancel, and truncated denial were classified as access.

Click/hover built `document.querySelector("")` when only `text` was set.

## Fix

After three not-found reads under the same directory prefix, or an incrementing numbered filename pattern, refuse further invented paths in that turn and point at `codedb list_dir`. A file that exists still reads. Root turns reset the tracker.

Native host names are resolved before MCP lookup.

Generic helper failures are `unavailable`, not access. Only an explicit Apple Event denial is `denied`.

Click/hover accept visible text (`innerText` / `aria-label` / placeholder). Renderer errors keep the underlying message.

Skipped open P1s that already have a PR, are already fixed on main, or need macOS/desktop design work are listed in the review notes.


justrach:
Merging onto the 0.0.302.2 line.

